PR #283, opened 2014-08-30
Less frustrating pop-up for new users

SomMeri:
Added "why isn't anything blocked" question with answer and link to FAQ into pop-up. The question is hidden if:
* no third party domains were detected,
* some domain is blocked,
* some domain was manually configured.

FFPB issue: https://github.com/EFForg/privacybadgerfirefox/issues/179

### DIFF
--- a/skin/popup.css
+++ b/skin/popup.css
@@ -168,6 +168,14 @@ font-size: 16px;
   margin: -7 0 0 59px;
   z-index: 20;
 }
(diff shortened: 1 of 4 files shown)


cooperq:
I think this still needs to happen and is long overdue. I am going to pull it and write some tests myself. Thanks for the patch @SomMeri !

PR #318, opened 2015-01-16
Remove cookie tests based on Opera test suite

gunesacar:
Privacy Badger's Travis builds have been failing for a while due to Opera test suite (testsuites.opera.com) which went offline a while ago. This branch removes the tests that depend on Opera's testsuite and add one simple test to make sure the cookies can be properly set by JS. The build passes with this patch:
 https://travis-ci.org/gunesacar/privacybadgerchrome/builds/47278699

It'd be nice to add more tests, of course.

### DIFF
--- a/tests/selenium/cookie_test.py
+++ b/tests/selenium/cookie_test.py
@@ -5,48 +5,18 @@ import unittest
 import pbtest
 
 
-class OperaCookieTest(pbtest.PBSeleniumTest):
-    """Run "Positive functionality" tests from Opera cookie tests.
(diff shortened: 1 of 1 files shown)


cooperq:
Yea it's unfortunate that these went away, apparently they are still available on github for self hosting and so we should probably set those up somewhere to make these tests work again. I have been meaning to do this but haven't gotten around to it, I think it's probably worth just removing them for now in the interest  of making the tests pass. I will work on that and merge this for now.

PR #321, opened 2015-01-21
Typofix in extractCookieString

nemobis:


cooperq:
Thanks for the patch! LGTM.

PR #327, opened 2015-02-06
Added options page with the ability to modify Whitelist and Filters

SwartzCr:
Sorry the diff on options.js is ultra-gnarly, I'd recommend just looking at the new file as the previous version of it was unused.


cooperq:
I think that this PR is invalidated by your other PR yes? I am going to close for now.

PR #357, opened 2015-04-16
Slider options

cooperq:


cooperq:
One issue with this for @SwartzCr 
When a slider in the options page is moved from green (Seen third parties) to userRed it does not stay there. When you reload the page that tracker is on, it moves back to green. I can demo this for you tomorrow if you want.

SwartzCr:
There seems to be an intermittent issue with FrequencyHeuristics.knownSubscriptions.seenThirdParties.parties being instantiated, a checking and instantiation function should be added as well

cooperq:
`FrequencyHeuristics.knownSubscriptions.seenThirdParties.parties` gets instantiated in the `migrate_version` function located at https://github.com/EFForg/privacybadgerchrome/blob/master/src/background.js#L107 

That function gets called when the version number is updated. This should be sufficient for upgrading users to the new version.

cooperq:
LGTM, other than a bunch of style nits which I think were artifacts from ABP. I will fix those.

PR #381, opened 2015-05-05
i18n

SwartzCr:
unused translation strings (from ABP) were removed from the locales file
more i18n support was added to strings in the html and JS
html strings are prefaced with the ID i18n_ and then the i18n script replaces them, js strings are gotten with a direct call to chrome's i18n module which reads from the locale files.
Some tooltips haven't yet been given i18n support, but I put in TODOs where it seemed apropriate


cooperq:
Need to add back in the name and description translations, but other than that this looks good. Still needs to be tested as well.

PR #390, opened 2015-05-20
Localstorage

SwartzCr:
please test this code, if it causes a performance hit it should be adjusted before merging

seen third party data is now being stored on disk instead of memory
Each third party origin gets its own filter and then in that filter ["subscriptions"][1] has an object that tracks the first parties it's been seen on

Added a check for first party domain in computeOriginsForTab.

added a bunch of comments about awful adblockplus code


SwartzCr:
It works for real now I promise @cooperq

cooperq:
lgtm

PR #397, opened 2015-05-27
Ui overhaul

cooperq:
Small updates to the chrome UI. Now at parity with firefox ui. Biggest update is that keys should now be centered and fixed so they don't go away when scrolling. 

CC @SwartzCr for review.


src/popup.js
```diff
@@ -205,7 +203,7 @@ function _badgerStatusTitle(action){
     block:        status_block,
     cookieblock:  status_cookieblock,
     noaction:     status_noaction
-  }
+  };
```
SwartzCr: why is this here?

SwartzCr:
Looks good on my machine, and works as expected.
If you wanted to spend a bit more time on this, I'd be a happy clam if you could get the vertical space between the domain name and the mouseover tooltip to be smaller, right now there's about a 20px gap between them, which makes the whole popup look awkward just to accomodate any given tooltip

src/popup.js
cooperq: It was there for DRYness before the i18n changes. Doesn't make sense any more. I will refactor this function.

cooperq:
Interesting, I don't see any weird gap on my machine, I wonder if it's related to your extreme resolution. Can you send me a screenshot?

Issue #402, opened 2015-05-30
social widget replacement double blocks certain widgets


SwartzCr:
go to: http://www.gamasutra.com/blogs/LarsDoucet/20131203/206159/On_Procedural_Death_Labyrinths.php
and notice two blocked google plus icons.
disable privacy badger and notice that there is only one google plus share button
when there should only be one replaced icon in the first place

gunesacar:
This is probably due to my [PR about social widgets](https://github.com/EFForg/privacybadgerchrome/pull/395/files#diff-a7e3669d2b1a8fab33514e1059d4cc30R77). Let me look into this.

SwartzCr:
fixed in #403

PR #403, opened 2015-05-30
Remove Google+1 selector which causes double blocking (#402).

gunesacar:
This commit should fix issue #402. The selector (`#___plusone_0`) removed in this commit caused PB to sometimes replace a Google Plus(+1) button with two replacement buttons.

At the heart of the issue is a strange race condition. Before [adding this selector](https://github.com/EFForg/privacybadgerchrome/pull/395/files#diff-a7e3669d2b1a8fab33514e1059d4cc30R77), sometimes (at least on my not-so-slow computer) +1 buttons had been rendered so quickly that PB couldn't replace them. This was because PB was looking for elements (`g:plusone` and `div.g-plusone`) that were already removed during the rendering of the +1 button.
This issue will again be visible after this commit (i.e. PB will sometimes miss +1 buttons) and I'll file a separate issue for that.
But (now I understand) it doesn't make sense to use the `#___plusone_0` selector, since it tries to find an already rendered +1 button's div (which is too late for replacing a social widget).

### DIFF
--- a/src/socialwidgets.json
+++ b/src/socialwidgets.json
@@ -73,8 +73,7 @@
 		"domain": "apis.google.com",
 		"buttonSelectors": [
 			"g\\:plusone",
-			".g-plusone",
-			"#___plusone_0"
+			".g-plusone"
 		],
 		"replacementButton": {
 			"details": "https://plusone.google.com/se/0/_/+1/fastbutton?url=",


SwartzCr:
This seems to no longer block the google plus button, on the site referenced in #402 (for me) the button shows up unperturbed

SwartzCr:
restarting my computer seems to have fixed it

PR #418, opened 2015-06-12
Remove the low entropy cookie names list

pde:
Since we don't use it, and there probably isn't a practical way to use it, and it may confuse people looking at the entropy calculation code.


cooperq:
Agreed, it was an experiment but deemed ineffective. Better to remove the cruft.

PR #422, opened 2015-06-12
Attempt to clarify Noah's correction :)

pde:
@SwartzCr is this what we mean?


cooperq:
Clarified a bit further.

PR #430, opened 2015-06-18
Detect potential localStorage supercookies.

gunesacar:
localStorage based potential supercookie detection, see #362


cooperq:
This looks great! I am going to merge, any further fixes that need to happen can be filed as bug reports I think.

Issue #405, opened 2015-06-01
Blocked Domain List not persisting between sessions


SwartzCr:
The list of all domains blocked by privacy badger should be stored so that their DNT policy can be checked daily to unblock them should their hosts agree to our DNT policy.
They seem to not persist between sessions, one way to see this is to load privacy badger, surf for a while, check chrome.storage.local.get('blockeddomainslist') and then restart chrome and check it again and notice a difference.
This storage should either be moved to localStorage (which bizarrely is a different thing) or to a adBlock style filter.

SwartzCr:
fixed in #417

PR #457, opened 2015-06-30
switched cookieblocklist to local synchronous storage

SwartzCr:
moved cookieblocklist from chrome.storage.local to localStorage, made it synchronous to fix the error where local storage gets wiped (as with #405)
Also fixed a line in the migration script for blocked domains list where it was trying to get the length of an object as opposed to the number of keys in it


cooperq:
Your tests appear to fail, can you fix the tests first?

cooperq:
LGTM

PR #459, opened 2015-07-02
Issue 446

cooperq:
do not mark things as trackers in the popup unless they are actually tracking on the current tab. cc @SwartzCr for review.


cooperq:
This can be tested by visiting e.g. cracked.com with third party cookie blocking turned off in the browser, thus marking facebook.com as a tracker, and then visiting buzzfeed.com with third party cookie blocking turned on in the browser. You should see facebook marked as not tracking you on buzzfeed.com but marked as tracking you on cracked.com

SwartzCr:
looks great!

PR #469, opened 2015-07-10
Restyle error reporting popup

taravancil:
In addition to restyling the popup, this adds a feature to notify the
user whether or not their report was submitted successfully.

### DIFF
--- a/_locales/en_US/messages.json
+++ b/_locales/en_US/messages.json
@@ -100,13 +100,28 @@
         "message": "Cancel"
     },
     "report_terms": {
-        "message": "This will send information to the EFF including: the site your on, your browser version, the version of Privacy Badger, and the state of all of the sliders statuses for this site"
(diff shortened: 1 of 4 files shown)


taravancil:
Let me know what you think of the new layout. Happy to make changes!

SwartzCr:
Looks good, two minor nit-picks and then I'm happy to merge
The success marker only gets reset when the whole window is closed, not when the overlay goes away, it'd be nice to add a $("#report_success").toggleClass("hidden"); to the overlay close

also, <span id="report_close" class="i18n_report_close"></span>, isn't a 'button' so the mouse cursor over it is the same as it would be over text, rather than a pointer, that'd be swell if that could be changed too.

Other than that, it looks great!
Once those changes are in I'm happy to merge, would you be willing to do this in Firefox as well, I can point you to where the overlay code is there if so

taravancil:
Ok, this should address the two nit-picks!

SwartzCr:
looks great!

PR #471, opened 2015-07-12
Yellowlist soundcloud's CDN

pde:
sndcdn has a lot of subdomains.

I see the following:

**a-v2.sndcdn.com**
**wis.sndcdn.com**
**ec-media.sndcdn.com**
i1.sndcdn.com
i2.sndcdn.com
i3.sndcdn.com
i4.sndcdn.com
style.sndcdn.com

Music seems to play correctly with just the bolded ones yellow, but it seems safest to yellowlist everything?


SwartzCr:
lgtm

PR #476, opened 2015-07-18
now top levels show instead of their subdomains

SwartzCr:
implemented to show how many subdomains are being blocked, could also show what they are (but it takes up too much space)
in a perfect world the revert button would revert the TLD, right now it probably does nothing?


src/popup.js
```diff
@@ -400,6 +400,26 @@ function makeSortable(domain){
...
+ * this is a terrible function that repeats
+ * a lot of the work that getAction does
+ * because getAction stores things in mysery
+ * land and there's no real way to get what's
+ * in the ABP filters without repeatedly
+ * querying them
+ */
+function getTopLevel(action, origin, tabId){
+    if (action == "usercookieblock"){
+      return backgroundPage.getDomainFromFilter(matcherStore.combinedMatcherStore.userYellow.matchesAny(origin, "SUBDOCUMENT", getHostForTab(tabId), true).text)
```
cooperq: nit: needs semicolons

src/popup.js
```diff
@@ -435,9 +456,26 @@ function refreshPopup(tabId) {
         nonTracking.push(origin);
         continue; 
     }
+    else {
+      if (action.includes("user")){
+        var prevOrigin = origin;
+        origin = getTopLevel(action, origin, tabId);
+        if (prevOrigin != origin){
+          if (compressedOrigins.hasOwnProperty(origin)){
+            compressedOrigins[origin]['subs'].push(prevOrigin.replace(origin, ''))
```
cooperq: style nit: semicolon

src/popup.js
```diff
@@ -435,9 +456,26 @@ function refreshPopup(tabId) {
...
+          compressedOrigins[origin] = {'action': action, 'subs':[prevOrigin.replace(origin, '')]};
+          continue;
+        }
+      }
+    }
     originCount++;
     printable = _addOriginHTML(origin, printable, action);
   }
+  for (key in compressedOrigins){
+    printable = _addOriginHTML( key+" ("+compressedOrigins[key]['subs'].length+" subdomains)", printable, compressedOrigins[key]['action'])
```
cooperq: style nit: semicolon.

cooperq:
Okay so I tested this and the UI part of it appears to work. There are a few problems though.
1. This appears to significantly slow down the popup appearing, it drags like hell if one of the domains is overridden at the TLD level. 
2. The undo button doesn't work (but you knew this)
3. Moving the slider on the TLD gives this interesting result: The setting doesn't change in the popup, but in the options page we now have two settings, the original one and the one that the user has changed. 
doubledouble

cooperq:
Interesting, I disabled and re-enabled privacy badger and the slowness and jankyness went away. Possible memory leak? I am going to keep an eye on this .

cooperq:
Also, I noticed that your branch is not up to date with master. I wonder if being merged with master would fix any problems.

SwartzCr:
yeah, it's definitely increased compute time, but it shouldn't be slower by that much

I'll merge Master in, and then try to see if I can get the actions to work sensibly

SwartzCr:
woo everything should be fixed now! (except your weird slow down issue)

cooperq:
looks great! Weirdly I got the slowdown again but it was fixed by disabling and re-enabling privacy badger. I'm gonna merge into master and see if I get any more problems.

PR #478, opened 2015-07-19
Remove redundant code related to enabled/disable button (#451).

gunesacar:
Remove redundant code related to enabled/disable button and move closeOverlay out of init. This is to fix #451.

PR #480, opened 2015-07-21
fixing some issues with gcommit 39d2413642a5d36c5157488c46c9fa05304caa41

SwartzCr:
code was removed that generated and reverted user override control in #478


cooperq:
good catch! :+1:

PR #490, opened 2015-07-25
Locale: Dutch translation

rvaneijk:
I added a Dutch translation (nl.properties) and corrected a typo in the English and German translations.

@pde, @cooperq


rvaneijk:
@pde, @cooperq If I need to change something to have the build pass the test, please let me know.
From what I can tell, it has something to do outside of the changes that I suggest.
Rob

gunesacar:
The failing tests should be related to the updated policy key name, just sent a pull request to fix it.

PR #493, opened 2015-07-25
locales: added Dutch translation

rvaneijk:
Retry after build failed and fix by @gunesacar


cooperq:
This is great, thanks!

PR #497, opened 2015-07-27
added closing of overlay and deactivation of submit buttons

SwartzCr:
fixed problem where someone could send a lot of junk error reports all at once polluting the DB


cooperq:
lgtm

PR #498, opened 2015-07-28
full height options page. Fixes #495

### LINKED ISSUE #495: options menu doesn't take up full screen height
same as a FF bug
when you look at the list of seen domains on the options page it's really small even if your screen is really tall, it should be taller

cooperq:
cc @SwartzCr


SwartzCr:
lgtm

PR #501, opened 2015-07-28
Stable

cooperq:


cooperq:
merging stable into master, there was one commit from @SwartzCr that wasn't in master.

PR #505, opened 2015-07-30
Issue 481

cooperq:
This fixes the slider not sliding back as well as a few other bugs, sliders should slide back to roughly appropriate positions now. 
There is still an edge case where if b.scorecardresearch.com is user-allowed then sb.scorecardresearch.com is considered a subdomain and is also user allowed, but I will open that as a separate issue. 

cc @SwartzCr for review.


SwartzCr:
looks good

PR #508, opened 2015-07-31
Get action rules fix

SwartzCr:
fixes the problem where if a parent's rule was checked after a subdomain's rule it overrode it
this has existed in the code forever (probably) now it's fixed by explicitly checking for direct subdomain match


cooperq:
lgtm

PR #517, opened 2015-08-04
Keep comic prompt until user actually sees it

taravancil:


SwartzCr:
I see you're checking to see if the user has scrolled down to the comic, unfortunately for me (and this might just have to do with my resolution) but I can't seem to ever get the comic to a place where this comparison triggers, meaning the prompt stays in my privacy badger pop-up forever

I was thinking more of something like triggering it on click of <a href="#" class="jcarousel-control-next" data-jcarouselcontrol="true">›</a>
so it goes away when someone starts paging through the comic?

taravancil:
That's strange! In any case, I changed it to trigger the prompt to go away if the user navigates through the comic in any capacity (keyboard arrow, click on arrow, or click on slide #).

SwartzCr:
Awesome! That looks great! I'm gonna look into why TravisCI is failing and then merge it
thanks so much!

SwartzCr:
okay all the same tests fail in mainline (*sigh*) so I'm going to merge this

gunesacar:
> Awesome! That looks great! I'm gonna look into why TravisCI is failing and then merge it
> thanks so much!

Let me send a PR for that.

PR #521, opened 2015-08-05
Don't check policy key name.

gunesacar:
Changes in the policy names causes some tests to fail, only check the policy hash length.
https://travis-ci.org/EFForg/privacybadgerchrome/builds/74160685
https://github.com/EFForg/privacybadgerchrome/pull/517#event-374608802


cooperq:
Thanks! I love it when the tests pass :smile_cat:

PR #559, opened 2015-08-15
Update of Swedish locale

Najoj:
Made changes to Swedish translation and added strings from en_US that were missing. This translation is based on those English text strings than the previous Swedish translations.


cooperq:
Looks great! Thanks!

PR #569, opened 2015-08-19
Issue 499

SwartzCr:
fix problems with twitch.tv


SwartzCr:
see https://github.com/EFForg/privacybadgerfirefox/issues/499

Issue #600, opened 2015-09-06
German translation to Long


wolffc:
The Message "The Domains below seem to Track you" is to Long in German
and it Wraps out of its Gray Block. i suggest changeing the the Gray Block to a Block of Variable Height. to handle Any Ammount of Text.

![privacy-badger](https://cloud.githubusercontent.com/assets/1393783/9707112/3e9ee8b0-5539-11e5-934c-0a26e915233b.PNG)

SwartzCr:
haha, seems reasonably bad
I guess we should either make the font size dynamic or the size of that div dynamic/larger
I'll try to do this tomorrow

wolffc:
I Just Provided a fix for this Isse See the Pull request https://github.com/EFForg/privacybadgerchrome/pull/607

SwartzCr:
Great! Thanks so much for the pull request. We're currently dealing with some re-licensing issues for PB so we don't want to take outside contributions until we have that figured out. As soon as that's dealt with we'll be happy to merge this, but for now we have to wait on it for a little bit.
I hope that's okay, and thanks so much for submitting the change!

wolffc:
Hi you can consider my Contribution as "Public Domain" i wave all my rights.
Therefore it should not be a Problem in any Licence Change in the Future.

i hardly Think a change from "20px" to "auto" is Protectable by any copyright anyway.

Regards chris

PR #607, opened 2015-09-15
[fixes: #600] "nonTrackers" Box is to small for German Translation 

wolffc:
The German Translation for "The Domains below Don't seem to Track you" so
so long that the the words wrap out of the gray box

solution: set the box container height to "auto"

see Issue:  https://github.com/EFForg/privacybadgerchrome/issues/600

### DIFF
--- a/skin/popup.css
+++ b/skin/popup.css
@@ -228,7 +228,7 @@ font-size: 16px;
   text-align: center;
   color: #f9f9f9;
   background: #555;
-  height: 20px;
+  height: auto;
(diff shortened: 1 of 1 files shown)


cooperq:
This looks great! I just need to make sure you are comfortable with this new license file https://github.com/EFForg/privacybadgerchrome/blob/master/LICENSE (it has changed since you opened your pull request)

wolffc:
Yes I am Fine with the new License Agreement.

PR #613, opened 2015-09-23
Update License file

pde:
- Clarify current license status
 - EFF is working to offer dual licenses to the Privacy Badger codebase; we
   would like to ensure that all contributors agree to that change


cooperq:
lgtm

Issue #610, opened 2015-09-22
Domain whitelisting not working


narcolepticinsomniac:
Privacy Badger started blocking all images from vk.me. The problem is that on any given page images are being pulled from a dozen different subdomains like: 
 (NSFW)
http://cs628520.vk.me/v628520704/14c41/bKD7V76L_Kw.jpg
or https://pp.vk.me/c624921/v624921684/3ef21/oHPpUuQh6_I.jpg
Every different page can have a dozen new subdomains. Whitelisting "vk.me" should allow them all but it does nothing.

cooperq:
How are you whitelisting it? If you go to the options page and move the slider for vk.me from red to green that should work to whitelist all subdomains of vk.me, the second tab "whitelisted domains" is for disabling privacy badger when that domain is the *first party*. Does that make sense?

narcolepticinsomniac:
I did think it would whitelist it as a third party also but I figured out that that isn't the case pretty quickly. That  really isn't the issue. I obviously didn't explain it clearly. There is no slider for "vk.me", just dozens of different subdomains and dozens of new ones on every different page. You can see what I mean on some pages on vk.com itself: http://vk.com/videos-62344632?section=all
The same method is used on lots of external sites who use vk as a free content host. I honestly searched for a while and couldn't find an example that was SFW.

cooperq:
If you click on the gear icon in the privacy badger menu and go to the list of all domains you don't see an entry for vk.me? Can you double check to make sure? (try using ctrl+f to search for it).

narcolepticinsomniac:
No, none in Devtools either. They seem to use it strictly to host images and the "sub" part of the domain is always a prefix so they are all treated as unique by PB. Pretty SFW example: http://vk.com/sashagrey.

cooperq:
So if vk.me resources are getting blocked but vk.me is not showing up in the options page then this is a bug. Nothing should ever be getting blocked if it doesn't show up in the list of all domains in the options page. It is also strange that vk.me got blocked for you in teh first place. for me all of the vk.me domains are marked as 'not tracking'

cooperq:
@SwartzCr should probably take a look at this.

SwartzCr:
[...]

PR #624, opened 2015-10-05
added wildcard subdomains to whitelist settings

SwartzCr:
fixes issue brought up in #610

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -190,7 +190,15 @@ var Utils = exports.Utils = {
     if(localStorage.disabledSites && JSON.parse(localStorage.disabledSites).length > 0){
       var sites = JSON.parse(localStorage.disabledSites);
       for(var i = 0; i < sites.length; i++){
-        if(sites[i] === origin){ return false; }
+        var site = sites[i];
+        if(site.startsWith("*")){
+          if(getBaseDomain(site) === getBaseDomain(origin)){
+            return false;
+          }
+        }
+        if(sites[i] === origin){
+          return false;
+        }
       }
     }
     return true;


SwartzCr:
for @cooperq to review
to test: run pb normal, whitelist *.taobao.com, visit world.taobao.com -- nothing happens
run with this change: see that world.taobao.com, or ju.taobao.com have pb deactivated on them

narcolepticinsomniac:
It's not all that clear from the comments above, but it seems like you're still referring to sites being whitelisted when you visit those particular sites. The problem is with whitelisting third-party content injected into other pages. I provided examples in the other thread. Basically some sites (vk.com included) pull dozens and dozens of images per page from "vk.me". Every url has its own random prefix like: http://097624.vk.me/whatever or http://850279.vk.me/whatever.
PB treats each one as a unique domain and the only option is to allow them individually which is not any kind of real option. No content is ever pulled from vk.me without a prefix so the slider to whitelist it never appears. If that was an option, I don't know whether or not it would affect all the prefixed subdomains.
First party wildcard whitetlisting is good, but there needs to be the same option for third-party content. 
Personally I don't understand why whitelisting a site as a first party doesn't encompass any content they're loading into other pages. You've already given them you're seal of approval, which is either because you trust them or because blocking them screws things up. Either way, whichever reason you have for doing so most likely applies to their third-party content as well.

SwartzCr:
@cooperq feel free to ignore this until I get the tests to pass
@narcolepticinsomniac Sorry, there was a different issue requesting wildcard whitelisting (though I can't find it).
As far as the feature you're asking for you should be able to toggle that in the options menu, where you will see the base domain for a given third party listed. If you set it to green it will apply to all subdomains.

narcolepticinsomniac:
I guess guess I have my answer as to whether or not whitelisting the base domain would affect those  with prefixes. It doesn't. It still treats them as unique. Damn.

cooperq:
@SwartzCr are you still working on this?

cooperq:
huh, the tests seem to pass on my local. re-running travis now. Looks good otherwise. Once travis passes I will merge.

Issue #594, opened 2015-08-31
PB dropdown is too short and narrow


maranomynet:
The drop-down list of 3rd party (tracker) domains is...

1. ...too narrow so domain-names get needlessly truncated and it becomes impossible to determine what 1st-level domain it belongs to.
2. ...too short so even a modest list of domains requires scrolling and multiple screen-shots when communicating with others.

Please make the dropdown flexible - or at least resizable.

cooperq:
We can't make it resizable, and I'm pretty sure we can't make it flexible, but I will look into it. 
Related: EFForg/privacybadgerfirefox#565

nextLane:
Hi, would it be a good idea to give the entire domain-name as a tooltip instead of the present tool-tips that essentially elaborates on meaning of green, yellow and red  levels.

cooperq:
Perhaps, yes!

PR #635, opened 2015-10-14
Tooltip shows complete domain name, issue 594

nextLane:
PB dropdown is too short and narrow #594

Changes in UI, now entire domain-name is showed as a tooltip thus solving the first problem of "dropdown being too narrow so domain-names get needlessly truncated and it becomes impossible to determine what 1st-level domain it belongs to."

### DIFF
--- a/src/popup.js
+++ b/src/popup.js
@@ -311,10 +311,14 @@ function _trim(str,max){
  */
 function _badgerStatusTitle(action , origin){ 
   
+  var status_block = i18n.getMessage("badger_status_block");
+  var status_cookieblock = i18n.getMessage("badger_status_cookieblock");
+  var status_noaction = i18n.getMessage("badger_status_noaction");
+
   var statusMap = {
-    block: "Blocked ",
-    cookieblock: "Blocked cookies from ",
-    noaction: "Allowed "
+    block:        status_block,
+    cookieblock:  status_cookieblock,
+    noaction:     status_noaction
   };
 
   return  statusMap[action] + origin ;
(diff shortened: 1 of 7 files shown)


src/popup.js
```diff
@@ -309,19 +309,15 @@ function _trim(str,max){
...
-  var statusMap = { 
-    block:        status_block,
-    cookieblock:  status_cookieblock,
-    noaction:     status_noaction
+function _badgerStatusTitle(action , origin){ 
+  
+  var statusMap = {
+    block: "Blocked ",
+    cookieblock: "Blocked cookies from ",
+    noaction: "Allowed "
```
SwartzCr: have these strings be loaded from i18n like the previous strings were

SwartzCr:
this is really close to being perfect, but there's one more thing I'd like you to do.
Notice that previously it had been getting the text for these actions from the i18n script - this is what allows privacy badger to be translated into other languages.
If you could add your text to the english text file _locales/en_US/messages.json much like the previous ones (badger_status_block, badger_status_cookieblock, and badger_status_noaction) were, with similar names
Additionally you should remove those strings from the messages.json file if they're no longer being used in the code.
Let me know if you have any questions about this, but once you do that this should be ready to merge!

nextLane:
Any other changes to be done?

SwartzCr:
looks good!
I'm not going to merge it right now, because this is a change that will break other language's support, and I want to make sure I have a plan with how to deal with that before putting it in master. But consider your work done!

cooperq:
This is great! Thank you! I think that perhaps the best way to deal with this for other locals is to replace the changed strings with the english versions and let volunteers translate them again. Or we could try to translate them ourselves using google translate.

cooperq:
Looks great! Thank you! :+1:

PR #641, opened 2015-10-22
First Run push button animated

Tahirih-Natang:
Essentially I have animated the orange button and then added an animated arrow pointing to the button. I think this makes this button obvious to prompt any user to click on it.


lib/options.js
```diff
@@ -11,10 +11,10 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
  *
- * You should have received a copy of the GNU General Public License
+ * You should have reccentralizeeived a copy of the GNU General Public License
```
cooperq: This looks like a typo

lib/options.js
```diff
@@ -265,8 +265,9 @@ function toggleEnabled() {
   refreshIconAndContextMenu(tab);
 }
 
-// ugly helpers: not to be used!
+// ugly helpers: not Ofto be used!
```
cooperq: another typo

lib/options.js
```diff
@@ -393,13 +400,33 @@ function refreshFilterPage() {
     // todo: gross hack, use templating framework
     var action = optionsGetAction(origin, originList);
     if(!action){ continue; }
+      ++cnt;
```
cooperq: It looks like you are increasing count a second time here, since this also hapens in addOriginHTML. Is there a reason for this?

lib/options.js
```diff
@@ -393,13 +400,33 @@ function refreshFilterPage() {
...
     printable = _addOriginHTML(origin, printable, action);
+      //console.log(printable);
   }
-  printable += "</div>"
+  printable.data += "</div>"
```
cooperq: I'm pretty sure printable is a string at this point, why are you adding to the `data` property of this string?

lib/options.js
```diff
@@ -513,7 +539,16 @@ function syncUISelections() {
...
+/*//This is added by me
+document.addEventListener('click', function(){
+    document.getElementByClassName(".deleteButtonClass").onclick = hideDiv();
+    idOfDeleteButton = 'deleteButton';
 
+});*/
+function hideDiv(){
+    //document.getElementById("domainDiv").setAttribute('hidden', 'true');
+    alert(this.id);
+}
```
cooperq: what is the purpose of this function?

lib/options.js
```diff
@@ -513,7 +539,16 @@ function syncUISelections() {
   console.log("Sync of userset options: " + JSON.stringify(settingsDict));
   syncSettingsDict(settingsDict);
 }
+/*//This is added by me
+document.addEventListener('click', function(){
+    document.getElementByClassName(".deleteButtonClass").onclick = hideDiv();
+    idOfDeleteButton = 'deleteButton';
 
+});*/
```
cooperq: Please delete instead of commenting out

lib/options.js
```diff
@@ -393,13 +400,33 @@ function refreshFilterPage() {
...
+        $("#"+idd).bind('click', function() {
+            var someId = $(this).attr('id');
+            var tmpId = someId.split('_');
+            
+            var finalId = tmpId[1];
+            
+            var idOfHiddenDiv = 'domainDiv_'+finalId;
+            $('#'+idOfHiddenDiv).remove();
+             console.log(someId);
+//             alert('a die ohhhhh'+ idd);
```
cooperq: Please delete instead of commenting out

lib/options.js
```diff
@@ -393,13 +400,33 @@ function refreshFilterPage() {
     // todo: gross hack, use templating framework
     var action = optionsGetAction(origin, originList);
     if(!action){ continue; }
+      ++cnt;
     printable = _addOriginHTML(origin, printable, action);
+      //console.log(printable);
```
cooperq: Please delete instead of commenting out

lib/options.js
```diff
@@ -278,9 +279,15 @@ function _addOriginHTML(origin, printable, action) {
...
-  return printable + '<div ' + classText + '" data-origin="' + origin + '" tooltip="' + _badgerStatusTitle(action) + '" data-original-action="' + action + '"><div class="origin" >' + _trim(origin,30) + '</div>' + _addToggleHtml(origin, action) + '<div class="honeybadgerPowered tooltip" tooltip="'+ feedTheBadgerTitle + '"></div><img class="tooltipArrow" src="/icons/badger-tb-arrow.png"><div class="tooltipContainer"></div></div>';
+  var item =  printable + '<div id="domainDiv_' + cnt + '"' + classText + '" data-origin="' + origin + '" tooltip="' + _badgerStatusTitle(action) + '" data-original-action="' + action + '"><div class="origin" >' + _trim(origin,30) + '</div>' + _addToggleHtml(origin, action) + '<button class= "deleteButtonClass" id=id_'+cnt+'>X</button> <div class="honeybadgerPowered tooltip" tooltip="'+ feedTheBadgerTitle + '"></div><img class="tooltipArrow" src="/icons/badger-tb-arrow.png"><div class="tooltipContainer"></div></div>';
+    //console.log(cnt);
+    cnt ++;
+    return item;
 }
+//This function runs at refresh automatically, something quite strange
+//function _doSomeStuff(){
+//    console.log('Ouch!');
+//}
```
cooperq: Please delete instead of commenting out

skin/popup.css
```diff
@@ -289,3 +324,4 @@ font-size: 16px;
 .hidden {
   display: none;
 }
+/*This has been added by me*/
```
cooperq: Please delete instead of commenting out

skin/popup.html
```diff
@@ -29,12 +29,19 @@
 <script type="text/javascript" src="/lib/jquery-ui/js/jquery-ui-1.10.4.custom.min.js"></script>
 <script type="text/javascript" src="/src/popup.js"></script>
 <script type="text/javascript" src="/lib/i18n.js"></script>
-
+<script src="http://h2.flashvortex.com/display.php?id=2_1445512564_43453_495_0_1204_85_10_2_20" type="text/javascript"></script>
```
cooperq: What does this script file do? Please import the script locally, we never want to link to a remote script from privacy badger (to do so would be a privacy disaster).

cooperq:
Thank you for the commits Tahirih! I have a couple of comments in addition to the minor stylistic nit picks up above. First of all I notice that your commit breaks the selenium tests, please fix that if you would. Secondly, it appears that the blue bouncing arrow sticks around after the user has clicked the orange box to make it go away. Probably a better UI would be for the blue arrow to go away as well after the user has clicked the box. 
Additionally I'm not totally clear on what your second commit is doing. It appears to be adding a button to remove domains from the list, this button looks like it would only remove them temporarily since the domains are only removed from the dom, and then the domain would be back the next time the user viewed the popup, is that correct? 
I'm a bit concerned that this would confuse users, who might expect the domain to go away for ever, or not be clear as to what is happening. Can you explain to me what you think would be a use case for this feature?

Issue #623, opened 2015-10-05
isPrivacyBadgerEnabled is writen poorly


SwartzCr:
Privacy Badger has the ability to turn itself off for certain sites. The list of these sites is stored in localStorage, meaning that they're a JSONified string.
For every page load Privacy Badger loads this string from localStorage, parses it as JSON and then iterates through all sites on that list to see if the current domain matches any of them.
This should be a smarter data structure that requires less load time (read from localStorage for each page load??) and easier search (maybe hash it on startup?)

josmcg:
I'll take a look at it!

sam5125:
A possible pruning can be achieved by adding one more layer of abstraction, in the JSON with reference to beginning alphabet of the given white-listed domain. In this way we need not to do an exhaustive search.

Tahirih-Natang:
@sam5125 what if the white-list is growing? Firstly I will like to propose a binary search algorithm to start with. Such that on will not have to go through the entire list before the can know if the domain is blacklisted or not.

sam5125:
@Tahirih-Natang  It can be even more meaningful if the list is growing as we will be able to abstract out the domains by name. Of course binary search is a great option. My opinion was to modify the data structure such that we can prune most of the results altogether. 
@SwartzCr It is just a suggestion but will that be of any use as far as optimization is concerned as it will highly depend on the amount of data(specifically domains) to be white-listed.

SwartzCr:
I think the problem right now is that it loads it from and writes it to disk for every check. If it was stored in memory as an object it would have constant time look-up time, which would be faster than any search.
While improving the search algorithm might help a little, it doesn't help that we're still reading it all from disk for every call.
So what we need is a way to make sure that it's always current in memory and then the read time will be greatly improved.

SwartzCr:
As far as how many domains will be in the whitelist, it will be relatively small, almost always under 100

josmcg:
@SwartzCr I agree. When is the utils object instantiated and when is it killed? I'm not super familiar with chrome apps so I'm not sure how this cycle works. Do apps reload at every page? If not we should parse the array at startup, and cast them to String objects so our `===` won't have to do that every time.

SwartzCr:
we have a background page object which runs all the time (you can look at it from the developer console) and see what it's data structures look like. I assume the start-up stuff would be in src/background.js but I haven't taken a look at it.

nextLane:
[...]

PR #644, opened 2015-10-26
Fixed #421 and implemented #623

### LINKED ISSUE #421: Update Facebook Share image
Privacy Badger needs a replacement image for Facebook Share social widget. The [current image](https://github.com/EFForg/privacybadgerchrome/commit/70cd1b4c4676441a901b0a7cff7246661702f7bb)  has been added as a temporary placeholder and is a copy of Facebook Like button. This may cause some confusion.

Kinnera323:


cooperq:
LGTM :+1:

Kinnera323:
@cooperq  Please review this code change and suggest improvements if any.

src/utils.js
```diff
@@ -210,12 +210,14 @@ var Utils = exports.Utils = {
    **/
   disablePrivacyBadgerForOrigin: function(origin){
     if(localStorage.disabledSites === undefined){
-      localStorage.disabledSites = JSON.stringify([origin]);
+      var disabledSites = new Object();
+      disabledSites[origin] = true;
+      localStorage.disabledSites = JSON.stringify(disabledSites);
```
cooperq: Each time somone disables privacy badger for a new site you would create an object {"newsite.com": true} and save that to localStorage.disabledSites, this would mean you could only have one disabled site at a time. This method should append to the disabled sites object not overwrite it. This could be solved by having an object in the global scope for utils.js that cached the contents of localStorage.disabledSites.

src/utils.js
```diff
@@ -187,10 +187,10 @@ var Utils = exports.Utils = {
    * @returns {Boolean} true if disabled
    **/
   isPrivacyBadgerEnabled: function(origin){
-    if(localStorage.disabledSites && JSON.parse(localStorage.disabledSites).length > 0){
+    if(localStorage.disabledSites){
       var sites = JSON.parse(localStorage.disabledSites);
-      for(var i = 0; i < sites.length; i++){
-        if(sites[i] === origin){ return false; }
+      if(sites[origin]){
+        return false;
```
cooperq: This is a good start, but it still causes a disk read every time we call `isPrivacyBadgerEnabled`. We can get even more of a speed gain by reading localStorage.disabledSites once at startup and storing that in an object in the global scope for utils.js. This would also give us an elegant solution to the issue below.

src/utils.js
```diff
@@ -230,9 +232,8 @@ var Utils = exports.Utils = {
       return;
     }
     var disabledSites = JSON.parse(localStorage.disabledSites);
-    var idx = disabledSites.indexOf(origin);
-    if(idx >= 0){
-      Utils.removeElementFromArray(disabledSites, idx);
+    if(disabledSites[origin]){
+      delete disabledSites[origin];
```
cooperq: Efficiency is less of a concern here since this function would only be called infrequently. Still we should probably use a global object here instead of serializing from localStorage each time.

cooperq:
@Kinnera323 this is an excellent start! Please see my comments above, but good work so far!

src/utils.js
Kinnera323: @cooperq  I am defining a new Object only when disabledSites is empty. When it already has some disabledSites it appends as you said.

src/utils.js
Kinnera323: We can get more speed by doing so. But according to that we should write onto localStorage only before the chrome shuts down. As of now, I have no idea how to do this. If there is any, could you suggest me?

src/utils.js
cooperq: Ah, I see now. You are correct.

src/utils.js
cooperq: We could happily write to localStorage each time we update the object in memory as long as we do it asynchronously.

Kinnera323:
globally loaded localStorage.disabledSites in memory, so that access time is saved! As you mentioned earlier.

cooperq:
This looks good, there is just one more step: for people who have already installed privacy badger and have some disabled sites we need to migrate the data structure from an array to an object where the keys are domains and the value is "true". If you add that I think this feature is complete. This is probably a good place to add such code: https://github.com/EFForg/privacybadgerchrome/blob/master/src/background.js#L109-L117

Kinnera323:
Please have a look at the updated code.Let me know in case of any issues!

cooperq:
This is looking good and it's almost there! Unfortunately your migration code will never run. This is because [this line](https://github.com/EFForg/privacybadgerchrome/pull/644/files#diff-961641dd6af66f5e3313307cf4eb8f1dR125) `Object.keys(sites)` will return the following for an array: `["0"]` (WTF Javascript‽) So you will need to come up with a better way to decide whether the object has already been migrated or not. You could check the version number or you could try to call certain methods to see if it is an array or not.

Issue #628, opened 2015-10-06
Consider making the red number be meaningfully gameable


pde:
Cecause Privacy Badger gets smarter over time, we could make the red "blocked" number in the UI be something that is strictly increasing as PB works better.

(This is a suggestion from a UI test... I'm *not* sure if there's a good way to do it)

SwartzCr:
Well notably the number could reflect the number of blocked trackers, as opposed to the number of noticed trackers. This wouldn't be hard to implement if that's what we decided to do

jawz101:
Take a look at this idea.  I'd have to draw it on paper I suppose https://github.com/EFForg/privacybadgerfirefox/issues/561

cooperq:
+1 I think that the way to do this is what @SwartzCr said. We could combine this with increasing the number of hits for a given third party domain, which could be shown in the options page.

cooperq:
fixed in #650

PR #650, opened 2015-11-01
#628 Making Red number gameable

sam5125:
The number now represents the total trackers blocked and not the potential trackers.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -783,6 +783,23 @@ function activelyBlockedOriginCount(tabId){
     }, 0);
 }
 
+/**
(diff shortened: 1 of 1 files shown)


cooperq:
This looks excellent! The only thing I would add is that we should probably increase the count for cookieblock and usercookieblock as well. If you add that I would be happy to accept this pull request!

sam5125:
@cooperq Please have a look at the changes. Now the number also include the domains with cookie block.

cooperq:
Great work! :+1:

Issue #651, opened 2015-11-02
Correction of type errors in popup.html


sam5125:
This is a possible typo with minimalistic effect on functionality. Replace stlye in the document by style.

cooperq:
fixed by @sam5125

PR #652, opened 2015-11-02
#651 fixed Corrected the type errors

sam5125:
The occurances of stlye is replaced by style.


cooperq:
looks good to me! :+1:

Issue #655, opened 2015-11-03
kaltura rule blocks video


terrorist96:
http://thefreethoughtproject.com/high-school-student-tackled-officers-phone-call-sick-mother/

Video is blocked by `cdnapi.kaltura.com`, `cdnbakmi.kaltura.com`, and `www.kaltura.com`. I understand it's a tracker, but I thought it should be brought to your attention.

cooperq:
I am unable to test this since playing the video requires flash. I wonder if this can be fixed by moving certain kaltura.com subdomains to cookieblocked. @terrorist96 can you test this and see what the minimum number of subdomains that would need to be added to the cookieblock list would be to fix this?

hansika:
@terrorist96 When I checked this URL, the domains were not blocked by Privacy Badger by default and I could play the video with no problem. But of course when I moved the sliders of the aforementioned domains to blocked state, the video got blocked.

terrorist96:
The following have to be moved to yellow to get the video to load:

cdnapi.kaltura.com
cdnbakmi.kaltura.com
www.kaltura.com

@hansika Yes, they weren't blocked for me initially either, but as I browsed around, and came back to that page, then it was blocked because PB realized it's a tracker.

hansika:
@terrorist96 yes. that's true. I sent a pull request after including the above domains in the yellowlist. The request has been merged.

terrorist96:
Thanks!

PR #659, opened 2015-11-06
issue #655

hansika:
I have added the cdnapi.kaltura.com, cdnbakmi.kaltura.com and www.kaltura.com domains to the sample_cookieblocklist.txt file


cooperq:
Thank you for the pull request. Can you please delete the swap file that got attached? Also, I would like to only add the bare minimum number of domains for video to work, can you confirm that all three of these domains are necessary or is there a smaller set of domains that could be allowed and still fix the problem?

hassaaa23:
Yes. I have tested with all possible subsets of the three domains. All three are required to play the video. Here's the modified commit.

hansika:
Yes. I have tested with all possible subsets of the three domains. All three are required to play the video. Here's the modified commit after deleting the swap file.

PR #660, opened 2015-11-06
QUnit_tests for heuristicblocking.js

Kinnera323:
Implemented unit tests for some of the functions.


Kinnera323:
@cooperq ,  Please have a look at the commented part of the code too.
For the commented code the tests fail because FilterStorage module should be executed before the tests code executes. (which is not happening now)

cooperq:
so two things here. one is that these tests appear to be failing on travisCI. Secondly please create a new pull request with just the tests you are trying to add (you still have the disabled sites code in here and the facebook share image, which are separate pull requests).

Issue #619, opened 2015-10-02
blocking netflix


cooperq:
in some cases privacy badger seems to block netlfix. This can be fixed by yellowlisting assets.nflxext.com

cooperq:
fixed in #662

PR #662, opened 2015-11-07
Issue #619

hansika:
This is the fix for issue #619. Added assets.nflxext.com domain into the yellow list.


cooperq:
please remove the .rhistory file. Looks good other than that. :+1:

Issue #663, opened 2015-11-07
prevent cookieblocked sites from accessing html5 local storage


cooperq:
we currently prevent sites on the cookieblock list from setting or reading cookies and referer(sic) headers. We should also prevent those sites from accessing html5 local storage. This can probably be done in a manor similar to what we do for blocking javascript cookies. see lib/clobbercookie.js

cooperq:
This is being worked on by @zmanian

PR #664, opened 2015-11-07
Block setting values to local storage if cookie blocking is enabled

zmanian:
Work in progress to #663 

Based of clobbercookie.js

If cookie blocking is enabled, it will clear Local Storage and thens block the api.

I'm not sure if it is possible to guarantee, that injected script runs before other scripts.

I have a test case here: https://github.com/zmanian/LocalStoragePrivacyBadgerTest

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -43,6 +43,7 @@
       "js": [
         "src/include.preload.js",
         "src/clobbercookie.js",
+        "src/clobberlocalstorage.js",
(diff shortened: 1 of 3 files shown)


zmanian:
@cooperq want feedback when you get a chance.

Do we need to had local storage behavior to the heuristics for this to be meaningful?

Does this handle the condition you intended to block?

PR #665, opened 2015-11-07
Have pointer cursor on switch and first run hover

mdamien:
Very small css fix but it's been bugging every time I open privacy badger :D


cooperq:
Awesome, thanks! :+1:

PR #667, opened 2015-11-08
Firstrun l10n

cooperq:
add slide text to localization strings (still need to actually implement, but now people can start translating.)


cooperq:
merging TBR. cc @jsha or @SwartzCr

Issue #585, opened 2015-08-25
First run page pointer isn't obvious enough


SwartzCr:
I've done some hallway testing, and continue to get screenshots where people haven't gotten rid of the 'learn how privacy badger is protecting your browser' dialogue. I'm afraid that people don't realize that they can get rid of it, leading to permanently reduced screen pop-up size.

My two initial thoughts on what to do are to have the orange button in the pop-up animate (maybe bounce) to draw attention to people to click it, and then to have a similarly animated arrow pointing at the 'next panel' button on the first run page.

I'm going to tag @taravancil in case she has opinions on this since she made the original

cooperq:
:+1:

taravancil:
I like that idea. It's too big to stay in the popup! What about animating it and also hiding it after the popup has been open a set number of times? Even with an animation, I imagine some people will *never* click it.

I'll do this over the weekend!

cooperq:
What about putting it on top of the rest of the window, and making it more apparent that you should click on it? Maybe this needs some A B testing.

SwartzCr:
@taravancil do you think you still have time to take a crack at this? Otherwise I might try to assign it to a prospective intern

Tahirih-Natang:
@SwartzCr I think I can take a look at this! But a good number of proposals have been made on how to enhance this feature. But what if we just add an image that carry the message "Click on me to Learn ........."

SwartzCr:
@Tahirih-Natang The current status is there is a large orange box that says that, but people aren't clicking on it! We want to make it more obvious and harder to ignore

Neha--Agarwal:
How about making a semi transparent layer over the popup for the first time when the extension is installed? Once the user clicks a button which says something like 'Got it', we can move the "learn text" in the way it is at present.

Tahirih-Natang:
Here:
https://github.com/Tahirih-Natang/privacybadgerchrome/commit/35dbb10c80cf3233450c17b7382697c431fac185

cooperq:
I think that maybe something like what @Neha--Agarwal suggests would be a good plan maybe with some modification, lets have a semi transparent layer over the popup that doesn't go away until the user clicks to see the slideshow (we don't have to show it at all if the user has already viewed the slideshow). I think that this could replace the orange box that's there now entirely.

PR #672, opened 2015-11-08
Tutorial overlay

tinkerb:
Fixed issue #585


cooperq:
Thanks @tinkerb! This is a good start but there are a couple of issues to address before I can accept this. 
1. There appears to be an extra file: `scripts/dummy-chromium.pem` please delete it.
2. The changes to the firstRun.html page: https://github.com/EFForg/privacybadgerchrome/pull/672/files#diff-f0a97cbe285a62f4f9b5d1bafd12558cR32 probably don't need to be there, also they make the first run page look broken. 
3. The overlay in the popup does not link to anything
4. when the "X" is clicked, the popup only temporarily goes away, reapearring the next time you open the popup. This should probably go away permanently. This could be done by checking the value of localStorage.seenComic before displaying it and setting that value to true once the link or the x are clicked. 
5. The overlay is a little ugly looking right now, it could use some CSS love. 

If this is too much work for you let me know! You have already done a lot of work here, I am happy to take what you have and finish it if you do not have the time.

tinkerb:
Hi Cooper,

I've actually been working on it a bit since this weekend. I realized there
were some issues, and most of what you've listed has been fixed and just
not committed at this point.

The thing I'm having an issue with is getting the value of seenComic to
change when someone clicks the x without being reset when it reloads. I'll
work on it a bit today though and may message you again if I can't get it
working.

Thanks for the response!

Lesley

cooperq:
superseded by #682

Issue #578, opened 2015-08-20
disqus comments


terrorist96:
a.disquscdn.com
uploads.disquscdn.com

needs to be defaulted to yellow.

PR #674, opened 2015-11-09
Issue#578

hansika:
This is the fix for issue #578. Added the two domains a.disquscdn.com and uploads.disquscdn into the yellowlist.


cooperq:
looks great! Thanks!

PR #680, opened 2015-11-11
isPrivacyBadgerEnabled made efficient #623

nextLane:
#623 
Converted disabledSites array to JavaScript object. 
Upgrade supported.

### DIFF
--- a/_locales/de/messages.json
+++ b/_locales/de/messages.json
@@ -3,13 +3,13 @@
         "message": "Domain hinzufügen"
     },
     "badger_status_block": {
-        "message": "Dieser Tracker ist BLOCKIERT. Schieben, um Blockierung aufzuheben oder Cookies zu blocken."
+        "message": "Gesperrt "
     },
     "badger_status_cookieblock": {
-        "message": "Die Cookies dieses Trackers sind blockiert. Schieben, um den Tracker zu blockieren oder komplett frei zu schalten."
+        "message": "Blockiert Cookies von "
     },
     "badger_status_noaction": {
-        "message": "Diese Domain ist nicht blockiert. Schieben, um komplett zu blocken oder Cookies zu blocken."
+        "message": "Dürfen "
     },
     "description": {
         "message": "Privacy Badger schützt sie vor Trackern während Sie im Web surfen !"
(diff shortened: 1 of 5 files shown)


src/utils.js
```diff
@@ -210,12 +213,14 @@ var Utils = exports.Utils = {
    **/
   disablePrivacyBadgerForOrigin: function(origin){
     if(localStorage.disabledSites === undefined){
-      localStorage.disabledSites = JSON.stringify([origin]);
+      var obj = {};
+      obj[origin] = true;
+      localStorage.disabledSites = JSON.stringify(obj);
```
kaushik94: You should never access `localStorage` naked. `setItem` and `getItem` are wrappers for this and they are the recommended way of doing. This is an excellent SO answer http://stackoverflow.com/questions/12632463/is-localstorage-getitemitem-better-than-localstorage-item-or-localstoragei/24904863#24904863.

src/utils.js
```diff
@@ -36,6 +36,7 @@
 require.scopes["utils"] = (function() {
   
 var exports = {};
+var disabledSites = localStorage.disabledSites;
```
kaushik94: `var disabledSites = localStorage.getItem("disabledSites");`

src/background.js
```diff
@@ -112,10 +112,25 @@ function migrateVersion(prevVersion,currentVersion){
...
 
+
+/**
+ * migrates disabled sites from array data structure to object
+ */
+function migrateDisabledSites() {
+  
+   var obj = {};
+   var sites = JSON.parse(localStorage.disabledSites);
+  for(var i = 0; i < sites.length; i++){
```
kaushik94: Here are a few better ways.
http://stackoverflow.com/questions/26264956/convert-object-array-to-hash-map-indexed-by-an-attribute-value-of-the-object
http://stackoverflow.com/questions/28835262/convert-array-to-hashmap-key-value-pair-javascript

nextLane:
Closed the PR for now, tests failed and the code needs some modifications. To address the issue of direct access to local storage as @kaushik94  rightly pointed out, opened the issue #685

PR #682, opened 2015-11-13
Tutorial overlay now stops showing up once a user has clicked x

tinkerb:
The overlay might need a little more content or css improvement, but the basic functionality is there.


cooperq:
This is looking better, there are a couple of things blocking me from accepting this currently:

1. There is still some extra content at the top of the first run page at: chrome-extension://bndnafjcebipcpjificpmjckmolooach/skin/firstRun.html
2. The overlay does not go away permanently when the X is clicked nor when the tutorial is viewed. Only the button to view the tutorial goes away. The button looks at a property `localStorage.seenComic` to decide whether or not it should display. The whole overlay should do the same. 
3. The text in the overlay should be added to the localization file instead of directly in popup.js so that it can be translated into other languages.

cooperq:
@tinkerb is this still something you are interested in working on?

tinkerb:
@cooperq I can't reproduce items 1 or 2. I believe they are fixed in the version I am running, which has now been pushed to my branch. Will work on 3.

cooperq:
Awesome! I will check it out again when you have the strings in translation. Thank you so much!

Issue #683, opened 2015-11-13
Poorly formed tooltip over sliders in User Filter Settings


nextLane:
Origin is not shown, presently, they appear as 'blocked', 'allowed', 'blocked cookies from'.

cooperq:
Thanks for finding and fixing this! :+1:

PR #684, opened 2015-11-13
Tooltip formed properly in options.js #683

nextLane:
#683 Patch to properly form tooltip texts responsible due to options.js


cooperq:
Excellent, thanks!

Issue #686, opened 2015-11-15
Image icons on Settings page remain fixed in position wrt the window. 


nextLane:
These neither scroll up/down or adjust when browser window is resized. 
![bugg](https://cloud.githubusercontent.com/assets/8895985/11166599/817b9fd0-8b61-11e5-9cc7-32cc8fccaabe.PNG)

cooperq:
LGTM :+1:

PR #687, opened 2015-11-15
Fixing the scrollability of image #686

nextLane:


cooperq:
looks good to me. fixes #686

PR #692, opened 2015-11-19
Netflix blocking

buren:
Sometimes privacy badger blocks Netlfix, seems to be the same issue as reported in #619 and fixed in #662


cooperq:
I think we should just add all of nflxext.com to the cookieblock list instead of just individual CDNs since more can always be added.

cooperq:
fixed in #696 thanks!

PR #705, opened 2015-12-08
Cookie test cases

PollyP:
This PR is a first shot at a PB selenium test. If it looks like I'm headed in the right direction with this, I'll start adding to this suite.

It implements a basic cookie test, where three different websites are loaded, all of which pull in a tracking cookie from a fourth site. The fourth site has no DNT policy. Each time the popup UI is checked to make sure the third party site is greenlisted. When the first site is reloaded, the popup UI is checked to make sure that the third party's site has been yellowlisted.

I think the question here is, where will the testbed websites live? Just to get things rolling, I put them in my AWS s3 buckets. I'd like some feedback about where they should really live and admins I should be working with to get these in place. Getting this question sorted out first would make building on these tests a lot easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - "2.7"
 install:
+  - sudo apt-get update
   - sudo apt-get -qq install xvfb python-virtualenv
   - npm install
 before_script:
--- a/setup_travis.sh
+++ b/setup_travis.sh
@@ -10,8 +10,14 @@ else
     elif [ $machine = "i686" ] ; then 
         bits="32"; 
     fi;
-    version="2.12"
+    version="2.20"
     wget -O /tmp/chromedriver.zip "https://chromedriver.storage.googleapis.com/$version/chromedriver_linux$bits.zip"
     sudo unzip /tmp/chromedriver.zip chromedriver -d /usr/local/bin/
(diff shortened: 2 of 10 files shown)


PollyP:
I folded in the work from PR #709 to update chromedriver.

The Travis CI failed on the supercookie check, which is unchanged. I'll look into this.

PollyP:
Now I can't reproduce the super cookie assert fail that came up in https://travis-ci.org/EFForg/privacybadgerchrome/builds/97356897. I tried this 6+ times in my own travis ci environment, too.

My best guess: (a) there was a transient network failure that meant that either the first party or third party didn't get loaded, so (b) no supercookie was detected, so (c) the third party wasn't stored in the super cookie domains object in localstorage, and therefore (d) the assert fired.

cooperq:
Excellent work! This code looks great, and I am super happy to have more tests! Thank you very much for the patches. :+1: 
Travis is weird for sure, but the tests seem to be working now so i am going to go for it and not worry about it. (in fact i think you might have fixed some failing travis tests here.)
As for where the sites can live, I think we can probably host them from EFF, which seems like the most sustainable solution. Can you email me directly so we can get something set up?

PollyP:
Awesome! This week is bad for me but I will ping you directly next week and
we can discuss the mechanics.

PR #709, opened 2015-12-16
Update chrome, chromedriver for travis

PollyP:
Regression tests run locally, but fail on Travis CI due to flakey Chromedriver behavior. This PR updates Chromedriver from v.2.12 to v.2.20. Because that version of Chromedriver won't run with whatever version of Chromium lives in the testbed, setup_travis.py installs the latest stable version of Chrome too.

Also: apt-get could not always find python-virtualenv. I added an apt-get update before in .travis.yml, which appears to solve the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - "2.7"
 install:
+  - sudo apt-get update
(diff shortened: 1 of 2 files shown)


cooperq:
Excellent! Thanks, this is a huge help!

PR #719, opened 2015-12-30
Whitelist private domains

rcutmore:
Updated to whitelist requests from private domains and clean up formatting of the affected function checkAction in webrequest.js. Let me know if you'd like to see any changes to this. If this looks good I can update the Firefox version to properly check for 127.0.0.0/8 addresses.


cooperq:
This looks fantastic, thank you!

cooperq:
Also my apologies for the delay in looking at this. Such a delay should not happen in the future.

PR #725, opened 2016-01-15
yellowlist seatme.yelp.com -- third party reservation service

pde:
Seen at eg, http://www.lolosf.com/reservations/


cooperq:
fixed instead in 7ff7b7b due to merge conflict.

Issue #726, opened 2016-01-18
console error being thrown on all page loads


getify:
In a recent release of chrome canary (49.0.2623.0 canary (64-bit)), I started seeing this error in my console for every single page load:

```
extensions::uncaught_exception_handler:8 Error in event handler for (unknown): TypeError: Cannot read property 'trackers' of undefined
    at chrome-extension://pkehgijcmpdhfbdbbnkijodmdjhbjlgp/src/socialwidgets.js:321:26
```

This happens even for pages where I have privacy badger disabled.

cooperq:
Does privacy badger seem to be breaking as well? Or are you just seeing an error?

getify:
I am not using the "social widgets" feature of privacy badger, so I suspect that's why it's not breaking for me. It would seem that this feature is likely broken given that its script is failing with this error.

gunesacar:
I think this happens only when the user disables the social widget replacement since the [webrequest.js doesn't return anything](https://github.com/EFForg/privacybadgerchrome/blob/950aad76b7ae51c513d709fffb89c89da12c1e2a/src/webrequest.js#L734-L737) to [callback in socialwidgets.js](https://github.com/EFForg/privacybadgerchrome/blob/950aad76b7ae51c513d709fffb89c89da12c1e2a/src/socialwidgets.js#L321) when the widget replacement is disabled.

I [added a check](https://github.com/EFForg/privacybadgerchrome/pull/747/files#diff-b26ca211fc638234c77450adaaf8fc5cR321) to make sure the response is not undefined.

But, then, there appears to be a bigger problem: social widget replacement code runs regardless of whether PB is deactivated or social widget replacement is disabled. I [added a check](https://github.com/EFForg/privacybadgerchrome/pull/747/files#diff-b26ca211fc638234c77450adaaf8fc5cR348) similar to ones we already have in [supercookie.js](https://github.com/EFForg/privacybadgerchrome/blob/950aad76b7ae51c513d709fffb89c89da12c1e2a/src/supercookie.js#L120) or [fingerprint.js](https://github.com/EFForg/privacybadgerchrome/blob/950aad76b7ae51c513d709fffb89c89da12c1e2a/src/fingerprinting.js#L288).

cooperq:
Thanks Gunes! Closed via #747

PR #747, opened 2016-02-15
Check if PB is activated and social widget replacement is enabled before replacing widgets.

gunesacar:
It seems the social widget replacement code does not check if PB is enabled for the current page and if widget replacement is enabled in the PB options.

This should add a check before injecting the content script that replaces the widgets. Also checks the `response` object in the social widget callback which causes the relevant issue #726.


cooperq:
looks great! thanks!

PR #748, opened 2016-02-16
Selenium test to run the pbtest.org test suite

PollyP:
Created a new class, pbtest_org_test.py, to run the pbtest.org acceptance tests, and report the passes, fails, and tests with 'undefined' results. It will assert out on any failed tests.

### DIFF
--- a/tests/selenium/pbtest_org.py
+++ b/tests/selenium/pbtest_org.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python
+# -*- coding: UTF-8 -*-
+
+import pbtest
+import sys
+import time
+import unittest
+
+import IPython
+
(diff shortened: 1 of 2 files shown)


cooperq:
This looks awesome! Of course the tests are failing because the tests on pbtest.org fail, which is exactly what we expect. I am going to work on fixing the bugs related to these tests and merge this once fixed.

PollyP:
Coolio. So will all the existing tests in the selenium directory be
redundant given the pbtest.org website? Should they get removed? Or do you
think they still service a purpose?

cooperq:
Good news! This has been merged into the weapi https://github.com/EFForg/privacybadgerchrome/tree/weapi branch and will be in the next release! Thanks for your excellent work @PollyP and sorry if it seemed like I had forgotten about you, but rest assured that I had not!

PR #751, opened 2016-02-19
Reduce the test fragility by waiting some time to make sure the dynamic content is loaded.

gunesacar:
Add some wait time on test pages with third-party frames to make sure they're loaded properly. Add a convenience method to load URLs.

Here's an example flaky test, failed after removing comments!
https://travis-ci.org/gunesacar/privacybadgerchrome/builds/110258308

### DIFF
--- a/tests/selenium/breakage_test.py
+++ b/tests/selenium/breakage_test.py
@@ -15,11 +15,11 @@ class Test(pbtest.PBSeleniumTest):
     add Reddit comments etc."""
 
     def test_should_load_eff_org(self):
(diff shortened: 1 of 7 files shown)


cooperq:
This is fantastic! I think I was just getting bit by this bug on some of my own pull requests as well!

gunesacar:
I checked the failing builds from the [CI build history](https://travis-ci.org/EFForg/privacybadgerchrome/builds/). It seems the QUnit tests timeout more often than the others, probably because the QUnit tests run after the page load. #753 tries to address that by introducing a longer timeout.

PR #754, opened 2016-02-25
avira.com is releaseing a more secure browser. It has Privacy Badger …

Thorsten-Sick:
…pre-installed. To be able to quick-fix broken pages we will have to be able to whitelist important but broken domains. For that our extension will communicated the domains to whitelist to Privacy Badger. API for that added here.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -458,6 +458,26 @@ function teardownCookieBlocking(domain){
   CookieBlockList.removeDomain(domain);
 }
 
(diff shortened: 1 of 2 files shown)


cooperq:
If we are going to add any more APIs to privacy badger it might make sense to put them in their own source file, but I Think that this is fine for now. The code looks solid. I'm glad that privacy badger is going to be in the scout browser by default!

Issue #691, opened 2015-11-18
User Filer Settings - sort, bulk select, bulk set


mkocikowski:
Privacy Badger Options --> User Filter Settings tab should have the ability to sort tracking domains (by color), and to bulk select / set domains ("set all green to red").

cooperq:
I disagree, I think that such features would only serve to clutter the UI and slow down the page. More options is almost never a good decision for UI. And giving users the ability to change domain settings in bulk will likely lead to people making mistakes and unblocking or blocking things they didn't plan to. 
I will not implement this feature.

mkocikowski:
I currently show 565 "potential tracking domains". The current UI displays only 7 of them at a time, sorted alphabetically. Without the ability to sort or search the list, it is, for me, very difficult to interact with this data. Is there a way to access this data through other means (other than the graphical UI)?

syeo66:
At least it would be great to see all the manually adjusted domains without having to skim through all the 993 entries in my current list. This should be the absolute minimum... without that Privacy Badger will soon be worthless for me.

cooperq:
Your point is well taken. I do agree that searching would at least be nice. And I can see how sorting user overrides to the top would be useful for some. I'm still not comfortable with bulk select / bulk set but I think that some amount of searching and sorting is useful. Removing this from wontfix.

KidPluto:
This one looks like it has been completed.  Thanks.

PR #755, opened 2016-02-27
Add domain searching on options page

rcutmore:
This ports domain searching on the option page from the Firefox version. Issue #691 is related but this doesn't completely resolve it.


cooperq:
This looks perfect! Thank you for doing this!

PR #759, opened 2016-03-09
Refactor options.js

rcutmore:
This refactors options.js by removing some duplicated code and using a cached copy of origins (refreshing when necessary). Just let me know if this needs any changes.


cooperq:
LGTM, thanks again for the excellent work!

PR #763, opened 2016-03-11
Refactor domain search on options page

rcutmore:
This ports changes from the Firefox version to add a short delay before filtering tracking domains on the options page while user types.

[Pull request](https://github.com/EFForg/privacybadgerfirefox/pull/745) for the same changes to the Firefox version.


cooperq:
awesome, thanks!

PR #766, opened 2016-03-15
Tutorialoverlay update

tinkerb:

PR #768, opened 2016-03-16
Remove DRY code in popup.js and options.js

rcutmore:
This is a start for consolidating some of the HTML utility functions in popup.js and options.js. For the most part I just cleaned the functions up without making major changes, and I added some tests.

I think there are about 12 other functions that I can investigate consolidating. I figured it was best to split the changes up instead of submitting one huge pull request.

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -22,6 +22,7 @@ for (var i = 0; i < imports.length; i++){
       window[imports[i]] = backgroundPage[imports[i]];
 }
 var Utils = require("utils").Utils;
+var htmlUtils = require("htmlutils").htmlUtils;
 var filterWhitelist = {}
 
 with(require("filterClasses"))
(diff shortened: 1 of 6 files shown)


cooperq:
This seems like a valuable change for sure.  It looks like there a couple of bugs in the popup with this patch however:
1. Sliders don't work (user override doesn't stick, colors dont change)
2. Tooltips are gone

rcutmore:
Thanks for reviewing, I'll take a closer look.

rcutmore:
@cooperq I'm seeing those issues on the current *master* branch but not on my *refactor-html-utils* branch for this PR nor my local branch for #763.

Is it possible it's from #766? Looking at [popup.css](https://github.com/tinkerb/privacybadgerchrome/blob/e871908d4036da9fcad815ea58b0941c9df45c61/skin/popup.css#L215) from that PR it looks like there were some unresolved merge conflicts? I don't have enough time to look closer at the moment so I'm not sure if there are other issues.

cooperq:
Yea, you are correct it looks like that PR introduced a number of problems. Sorry! My fault! I am going to have to go over it and fix a bunch of bugs. I should have looked at it closer.

PR #769, opened 2016-03-16
add bandcamp cdn to cookie block list

SwartzCr:
bcbits.com seems to be bandcamp's CDN and is getting blocked for me. I had manually overridden some of the domains it uses, but it uses random subdomains p1.bcbits.com, f1.bcbits.com, s1.bcbits.com which mean that occasionally a bandcamp site will still randomly break.


cooperq:
lgtm

PR #783, opened 2016-03-21
Add Fasterize CDN to cookie block list

stefounet:
see https://github.com/EFForg/privacybadgerchrome/issues/774


cooperq:
Thank you!

PR #786, opened 2016-03-23
Fix issues with tutorial overlay changes

rcutmore:
This fixes some of the issues with #766. It looks like including *firstRun.js* in *popup.html* was causing issues with the tooltips and sliders. There are still other issues because the trackers link in the popup isn't fixed yet and the merge conflict in *popup.css* needs to be cleaned up.

### DIFF
--- a/skin/popup.css
+++ b/skin/popup.css
@@ -226,30 +226,17 @@ font-size: 16px;
   position: fixed;
   z-index: 100;
   background-color: #000;
-<<<<<<< HEAD
-  opacity: .85;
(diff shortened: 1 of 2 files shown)


rcutmore:
I cleaned up *popup.css* so I believe the only remaining issue is the trackers link. It's possible it wasn't working before the recent tutorial overlay changes though.

cooperq:
Indeed I think that the trackers link being broken is because the privacy badger website got redesigned. I will fix that.

PR #797, opened 2016-04-02
UI Stuff

cooperq:
* More intrusive tutorial nag popup, also easier to get rid of
* Add EFF Donate button

cc @jsha @Hainish @SwartzCr


SwartzCr:
lgtm!

PR #800, opened 2016-04-03
Fix crash when closing options page

### LINKED ISSUE #799: Crash after changing user filter settings on options page and closing tab
I'm having Privacy Badger crash reliably when changing the blocked status of a tracking domain on the options page and then closing the options page.

Steps to reproduce:

1. Open Privacy Badger's options page.
2. Change one of the sliders for a tracking domain.
3. Close options page tab.

It looks like there are a few possible issues. No changes are saved until the options page is closed (probably for performance reasons). When the options page is closed it triggers multiple listeners, each of which save the changes and refresh the tracking domain list on the user filter tab. The user filter tab refresh is triggered at least three times when closing the tab.

I'm going to take a look at this today and see if I can come up with a fix.

Potentially related issues:

- #791 
- #643

rcutmore:
This fixes #799. Changes are saved as soon as they are made instead of waiting until the options page is closed. There are further improvements that can be made but I wanted to submit this now as I'll be busy the next couple weeks.

This also fixes an unrelated issue where a change is made to an origin, the list of origins are filtered, and then closing the options page doesn't save the change. This is because the changes to save are determined by checking the displayed list of origins.

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -165,10 +165,8 @@ function unloadOptions()
   FilterNotifier.removeListener(onFilterChange);
 }
 
-function onFilterChange(action, item, param1, param2)
-{
-    syncUISelections();
-    refreshFilterPage();
+function onFilterChange(action, item, param1, param2) {
+  syncSettings();
 }
 
 function addWhitelistDomain(event)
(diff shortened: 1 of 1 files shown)


cooperq:
awesome thank you!

PR #807, opened 2016-04-12
Blockingtest

cooperq:
cc @SwartzCr for review. Can be tested by seeing that privacy badger still works and that the number of instances of "waiting for privacy badger" has decreased...


SwartzCr:
lgtm, doesn't look like it's going to remake the problem of wiping the data-store

PR #814, opened 2016-04-23
Use Ubuntu 14.04 for Travis tests

gunesacar:
Travis tests started to fail as Chrome doesn't support Ubuntu 12.04 anymore.

Use 14;04 machines for the Travis tests.

Install Chrome using Travis APT addon.

Read latest version of ChromeDriver from its release page.
We don't need to manually the bump the version again.

Remove redundant check about the machine architecture.
Travis machines are always 64 bit.

Do not install xvfb (already installed on Travis machines)

Run py.test with -v (verbose) switch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,21 @@
+sudo: required
 language: python
+os: linux
+dist: trusty
 python:
   - "2.7"
+addons:
+  apt:
+    sources:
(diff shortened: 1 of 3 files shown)


cooperq:
@gunesacar you are a hero! Thank you for fixing this :smile_cat:

PR #815, opened 2016-04-23
Add origin removal capability to options page

### LINKED ISSUE #589: "User Filter Settings" is unclearable
I can't seem to remove domains from "User Filter Settings", which is, ironically, very non-privacy respecting:

screen shot 2015-08-27 at 10 34 43 am

rcutmore:
This fixes #589 by adding elements for removing origins on the *User Filter Settings* tab of the options page. Origins are removed from the ABP filters (heuristic and user overrides) as well as from list of seen origins. I believe this covers everything but let me know if I'm missing something.

Also, let me know if you'd prefer these changes to be implemented after the switch to the new storage engine.

### DIFF
--- a/_locales/en_US/messages.json
+++ b/_locales/en_US/messages.json
@@ -99,6 +99,9 @@
         "description": "This is the label for the 'Remove selected' buttons.",
         "message": "Remove selected"
     },
+    "remove_origin_confirmation": {
+        "message": "Are you sure you want to remove this origin from Privacy Badger?"
+    },
     "report_broken_site": {
         "message": "Report Broken Site"
     },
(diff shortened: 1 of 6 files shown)


cooperq:
So, I think I am about a week or two away from releasing a version without adblock plus in it any more. The upside is that it will be so much easier to work on privacy badger once we get rid of ABP. The downside is that this code would have to be completely reimplimented for the new storage engine. 
I'm probably not going to do a new release for chrome before I release the version without ABP, so it might be best to hold off on this change or make the change against the weapi branch.

rcutmore:
No problem, I wasn't sure how far off the switch would be. I'll close this and either look at the weapi branch or wait until the switch.

PR #819, opened 2016-04-29
Get rid of Adblock Plus

cooperq:
Here it is. The code has been entirely refactored to get rid of adblock plus. All storage is handled by storage.js now. Other files have been refactored to rely on storage.js instead of ABP. 
The only remaining ABP code lives in background.js, this is because we will need it to migrate data from old versions of privacy badger. Once everyone is on the new version we will be able to remove the final vestiges of adblock plus code. 

This probably doesn't work on firefox yet, so please test on chrome. I reccomend running it with ./scripts/just_run.sh which will run it in a fresh chrome profile. 

CC @SwartzCr and @rcutmore for review, testing, and bugfixes.


doc/sample_cookieblocklist.txt
```diff
@@ -1,540 +1,540 @@
...
+ziplist.com
+zlcdn.com
+zombaio.com
+zope.net
+zopim.com
+zvents.com
+assets.nflxext.com
+assets.ubuntu.com
+y3.analytics.yahoo.com
+seatme.yelp.com
```
SwartzCr: yes! finally!

SwartzCr:
I'm getting a lot of errors when I run it from a clean browser:
```￼Unrecognized manifest key 'applications'.
  "applications": {
    "gecko": {
      "id": "privacybadger@mozilla.org"
    }
  },```

`￼This extension includes the key file '/home/swartzcr/Documents/coding/privacybadgerchrome/scripts/dummy-chromium.pem'. You probably don't want to do that.`

and an error that it can't load some adblockplus datastructure
```lib/compat.js:77 (Components.utils.reportError)
lib/adblockplus.js:1537 (anonymous function)
lib/io.js:59 (require.scopes.io.exports.IO.readFromFile)
lib/adblockplus.js:1529 (anonymous function)
lib/adblockplus.js:1614 (anonymous function)
lib/adblockplus.js:1633 (callback)
```

resulting in the plugin not working

src/background.js
```diff
@@ -34,29 +36,371 @@
...
+        for (var j = 0; j < windows[i].tabs.length; j++) {
+          refreshIconAndContextMenu(windows[i].tabs[j]);
+        }
+      }
+    });
+    pb.showFirstRunPage();
+    
+    pb.INITIALIZED = true;
+    console.log('privacy badger is ready to rock');
+    console.log('set pb.DEBUG=1 to view console messages');
```
SwartzCr: shoudl debug be set to false here? it seems to be set to true up at line 75

src/background.js
```diff
@@ -34,29 +36,371 @@
...
+      var cookieblock_list = pb.storage.getBadgerStorageObject('cookieblock_list');
+      var action_map = pb.storage.getBadgerStorageObject('action_map');
+
+      var newCbDomains = _.map(response.split("\n"), function(d){ return d.trim();});
+      var oldCbDomains = Object.keys(cookieblock_list.getItemClones());
+
+      var addedDomains = _.difference(newCbDomains, oldCbDomains);
+      var removedDomains = _.difference(oldCbDomains, newCbDomains);
+      pb.log('adding to cookie blocklist:', addedDomains);
+      pb.log('removing from cookie blocklist:', removedDomains);
```
SwartzCr: should this be formatted in some way? in python it'd be something like ", ".join(removedDomains)

src/storage.js
```diff
@@ -0,0 +1,432 @@
...
+};
+
+var getScore = function(action){
+  switch(action){
+    case pb.NO_TRACKING: return 0;
+    case pb.ALLOW: return 1;
+    case pb.BLOCK: return 2;
+    case pb.COOKIEBLOCK: return 3;
+    case pb.DNT: return 4;
+    default: return 5;
```
SwartzCr: is this just so you don't have to write out all three user actions?

rcutmore:
I'm not seeing the same errors as @SwartzCr when running the extension. I haven't run from a clean browser though, just uninstalled the current version and loaded this one, so I'll play around more to try replicating those issues.

These are some issues I'm seeing:

- I've seen the following error a few times while browsing: `extensions::uncaught_exception_handler:8 Error in event handler for webNavigation.onCreatedNavigationTarget: ReferenceError: getFrameUrl is not defined at chrome-extension://eokjccgnnbaiconeacneohdjopdcgjma/lib/popupBlocker.js:27:21`

- On options page, overriding action for a base domain sets the same override for all subdomains (ex: overriding *facebook.com* also sets override for *www.facebook.com* and *graph.facebook.com*). This seems intentional.

pb2

Clicking undo for base domain only updates base domain visually (as you'll see further below the subdomains are reset as well though).

pb3

Filtering/unfiltering causes all entries to display with previous user override setting.

pb4

However, re-opening the options page shows all entries with user override removed.

pb5

- On options page, undoing user override for subdomain has no effect if base domain has user override. This seems intentional but could possibly confuse users?

- On options page, can't turn off "Show count of blocked items" setting. When you uncheck setting and reopen options page, it's checked again.

- On options page, overriding action while origin list is filtered resets filter (i.e. all origins are displayed). This is an issue in the current version and I believe it's fixed in a [PR](https://github.com/EFForg/privacybadgerchrome/pull/815) that needs to be reworked for these changes.

I'll spend some more time reviewing and see if I can submit fixes for some of these.

src/background.js
cooperq: Indeed I have had it set to true for development purposes but we should set it back to false before we release.

src/background.js
cooperq: Nope, console.log automatically typecasts everything to a string.

src/storage.js
cooperq: yea, exactly. My assumption is that we will never get a non existent action through here. Though it might be safer to rewrite this to not make that assumption.

doc/DESIGN-AND-ROADMAP.md
```diff
@@ -61,6 +61,61 @@ Privacy Badger:
...
+action_map = { 'google.com': blocked, 'fonts.google.com', 'cookieblocked', 'apis.fonts.google.com': 'user_cookieblock', 'foo.tracker.net': 'allow', 'tracker.net': 'DNT', }
+snitch_map = {google.com: array('cooperq.com', 'noah.com', 'eff.org'), tracker.net: array(a.com, b.com, c.com)} 
+dnt_domains = array('tracker.net', 'dnt.eff.org')
+settings = {social_widgets = true, ...}
+cookie_block_list = "{'fonts.google.com': true, 'maps.google.com', true}"
+
+
+On Request():
+  check if privacy badger is enabled for this fqdn, if not then return
+  check if the fqdn is third party, if not then return
```
SwartzCr: does this mean PB won't block third parties on which PB is disabled?
So like if I disable pb on twitter and then there's an embedded twitter thing will it load or will it get blocked (assuming it's hueristic blocked)?

doc/DESIGN-AND-ROADMAP.md
```diff
@@ -61,6 +61,61 @@ Privacy Badger:
...
+  related_domains = array()
+  best_action == 'noaction'
+
+  for $domain in range(fqdn ... etld+1)
+    if action_map contains $domain
+      related_domains.push($domain)
+      
+
+    for each domain in related domains
+      if(score(domain.action) > score(best_action)
```
SwartzCr: what happens if twitter.com is user blocked, and gifs.twitter.com is user allowed or user cookie blocked? This implies that it gets blocked

src/background.js
```diff
@@ -34,158 +36,394 @@
...
 var DomainExceptions = require("domainExceptions").DomainExceptions;
 var HeuristicBlocking = require("heuristicblocking");
 var SocialWidgetLoader = require("socialwidgetloader");
-
-// Load social widgets
+var pbStorage = require("storage");
+var webrequest = require("webrequest");
 var SocialWidgetList = SocialWidgetLoader.loadSocialWidgetsFromFile("src/socialwidgets.json");
+var Migrations = require("migrations").Migrations;
+  var FilterStorage = require("filterStorage").FilterStorage;
```
SwartzCr: unindent

doc/DESIGN-AND-ROADMAP.md
cooperq: push in this case is actually shift (add to the front of the array) and is so in the code, which makes it so that the array is ordered least to most specific which means that this works as expected (a user rule for gifs.twitter.com overrules twitter.com) but I will update the docs to reflect that.

doc/DESIGN-AND-ROADMAP.md
cooperq: Nope just unclear docs. I mean to say 'check if pb is enabled for the tab host'

src/background.js
```diff
@@ -34,158 +36,394 @@
...
+          pb.initializeCookieBlockList();
+          pb.initializeDNT();
+        }
+      });
+
+      // Show icon as page action for all tabs that already exist
+      chrome.windows.getAll({populate: true}, function(windows) {
+        for (var i = 0; i < windows.length; i++) {
+          for (var j = 0; j < windows[i].tabs.length; j++) {
+            refreshIconAndContextMenu(windows[i].tabs[j]);
```
SwartzCr: let's say I have 100 tabs open and I install Privacy Badger, will this refresh all of my tabs?

src/background.js
```diff
@@ -34,158 +36,394 @@
...
+      });
+
+      // Add any new cookie block domains who's parent domain is already blocked
+      _.each(addedDomains, function(domain){
+        cookieblock_list.setItem(domain, true);
+        var baseDomain = window.getBaseDomain(domain);
+        if(action_map.hasItem(baseDomain) &&
+           _.contains([pb.BLOCK, pb.COOKIEBLOCK],
+                      action_map.getItem(baseDomain).heuristicAction)){
+          pb.storage.setupHeuristicAction(domain, pb.COOKIEBLOCK);
```
SwartzCr: so only things that are blocked get cookie blocked
so the flow is that if a domain isn't seen, we wait til it gets blocked and then the next day the cookie block list fetch sets it to cookieblock?

doc/DESIGN-AND-ROADMAP.md
```diff
@@ -61,6 +61,61 @@ Privacy Badger:
 
 #### Further Details
 
+Datastructures:
+
+action_map = { 'google.com': blocked, 'fonts.google.com', 'cookieblocked', 'apis.fonts.google.com': 'user_cookieblock', 'foo.tracker.net': 'allow', 'tracker.net': 'DNT', }
```
SwartzCr: there is a typo in this line

src/background.js
```diff
@@ -206,241 +444,61 @@ function getAction(tabId, origin) {
...
-  // Ignore fragment identifier
-  var index = url.indexOf("#");
-  if (index >= 0) {
-    url = url.substring(0, index);
+ * @return {Boolean} true if the url is allowed false if not
+ */
+function isWhitelisted(url) {
+  var host = window.extractHostFromURL(url);
+  var action_map = pbStorage.getBadgerStorageObject('action_map');
+  var action = action_map.getItem(host);
```
SwartzCr: shouldn't this be get best action?

src/background.js
```diff
@@ -884,77 +603,94 @@ function updateCount(details){
...
+  while (parentData)
+  {
+    var frameData = parentData;
+
+    parent = frameData.parent;
+    parentData = webrequest.getFrameData(tabId, parent);
+
+    var frameUrl = frameData.url;
+    var parentUrl = (parentData ? parentData.url : frameUrl);
+    if ("keyException" in frameData || isWhitelisted(frameUrl)){
```
SwartzCr: whitelisted checks action_map, shouldn't it be kept somewhere else?

src/fingerprinting.js
```diff
@@ -73,7 +73,7 @@ function getPageScript() {
           detail: messages
         }));
 
-        // clear the queue
+        // clear the queue
```
SwartzCr: lint

src/background.js
cooperq: This does not refresh the tab, just makes sure that icon gets set right.

src/background.js
cooperq: No, we cookieblock things when they are blocked as well. This is saying when you get the cookie block list, for any new domains on the list, if their base domain is blocked, set up cookie blocking for the new domain.

doc/DESIGN-AND-ROADMAP.md
cooperq: where?

src/background.js
cooperq: Yup! I have fixed this.

src/background.js
cooperq: Where else would we keep it? This is checking if a domain is unblocked essentially, not if privacy badger is disabled.

src/options.js
```diff
@@ -14,37 +14,19 @@
  * You should have received a copy of the GNU General Public License
  * along with Adblock Plus.  If not, see <http://www.gnu.org/licenses/>.
  */
+ // TODO: This code is a hideous mess and desperately needs to be refactored and cleaned up.
```
SwartzCr: tough but fair

src/options.js
```diff
@@ -198,51 +164,16 @@ function removeWhitelistDomain(event)
...
-    if (! origins[origin]) {
-      origins[origin] = 'noaction';
-    }
+function getOrigins() {
+  var origins = {};
+  var action_map = pb.storage.getBadgerStorageObject('action_map');
+  for (var domain in action_map.getItemClones()) {
+      var action = pb.storage.getBestAction(domain);
+      // Do not show non tracking origins
+      if(action != pb.NO_TRACKING){
```
SwartzCr: so we're not giving people the option of overriding non-tracking domains on the options page?

src/popup.js
```diff
@@ -16,58 +16,16 @@
  * You should have received a copy of the GNU General Public License
```
SwartzCr: it'd be cool if all of this code could be shared between options and popup
refactor 2.0

src/storage.js
```diff
@@ -0,0 +1,430 @@
...
+  }
+};
+
+/**
+ * get the current presumed action for a specific fqdn, ignoring any rules for subdomains
+ * below or above it
+ * @param {Object} domain domain object from action_map
+ * @returns {String} the presumed action for this FQDN
+ **/
+var getActionForFqdn = function(domain){
```
SwartzCr: I'm really confused now about when we'd ever be querying values directly from the action map, or what the action map is used for if we can call this function

src/storage.js
```diff
@@ -0,0 +1,430 @@
...
+      // Order from base domain to FQDN
+      relevantDomains.unshift(action_map.getItem(subdomains[i]));
+    }
+  }
+
+  // Loop through each subdomain we have a rule for from least to most specific
+  // and keep the one which has the best score. 
+  for( i = 0; i < relevantDomains.length; i++ ){
+    var action = getActionForFqdn(relevantDomains[i]);
+    if(getScore(action) >= getScore(best_action)){
```
SwartzCr: yes good

src/webrequest.js
```diff
@@ -223,52 +115,52 @@ function onBeforeRequest(details){
...
+    // to block
+    if (requestAction == pb.BLOCK) {
+      // Notify the content script...
+      var msg = {
+        replaceSocialWidget: true,
+        trackerDomain: window.extractHostFromURL(details.url)
+      };
+      chrome.tabs.sendMessage(details.tabId, msg);
 
-    if (requestAction == "cookieblock" || requestAction == "usercookieblock") {
```
SwartzCr: is this a change in when social widgets get blocked?

SwartzCr:
I'd like to see @jsha or bill look over the badgerStoarage stuff

src/options.js
cooperq: I mean, it's a hideous mess because popup.js was a hideous mess. Not because of anything you did :smile_cat:

src/options.js
cooperq: Yea, I felt like it cluttered things too much? Although that may not be the right call. I am willing to consider changing this behavior.

src/popup.js
cooperq: :+1: yass. I think that @rcutmore started doing some of that work by moving things to htmlUtils.js but obvs this is a WIP

src/storage.js
cooperq: If say we wanted to check if a domain was even in the action map at all, or if we wanted to get a list of all the domains that are in action map.

src/storage.js
cooperq: :snowman: :+1:

src/webrequest.js
cooperq: nope! This line was replaced with pb.COOKIEBLOCK and so forth. This shouldn't be a change in when social widget replacement happens.

cooperq:
@jsha or @Hainish do you have time to review storage.js for me on monday morning?

Hainish:
@cooperq I should have some time on Monday.

doc/DESIGN-AND-ROADMAP.md
rcutmore: Looks like the issue is with `'fonts.google.com', 'cookieblocked'`. I submitted a PR to fix this and reformat that section of the docs a bit.

src/storage.js
```diff
@@ -0,0 +1,430 @@
...
+BadgerStorage.prototype = {
+
+  /**
+   * check if this storage object has an item 
+   * @param {String} key the key for the item
+   * @return boolean 
+   **/
+  hasItem: function(key){
+    var self = this;
+    return !!self._store.hasOwnProperty(key);
```
rcutmore: I'm a bit of a JS newb, is there a need to coerce the result of *hasOwnProperty* to a boolean? Is there a situation where the result would be something other than a boolean?

src/storage.js
```diff
@@ -0,0 +1,430 @@
...
+var setupHeuristicAction = function(domain, action){
+  _setupDomainAction(domain, action, "heuristicAction");
+};
+
+/**
+ * Sets up a domain for DNT
+ * @param {String} domain Domain to add
+ */
+var setupDNT = function(domain){
+  _setupDomainAction(domain, "true", "dnt");
```
rcutmore: Should this set a boolean rather than a string? `_setupDomainAction(domain, true, "dnt");`

src/storage.js
```diff
@@ -0,0 +1,430 @@
...
+var setupDNT = function(domain){
+  _setupDomainAction(domain, "true", "dnt"); 
+};
+  
+/**
+* remove DNT setting from a domain
+* @param domain FQDN string
+**/
+var revertDNT = function(domain){
+  _setupDomainAction(domain, null, "dnt");
```
rcutmore: The action_map comments [near the top](https://github.com/EFForg/privacybadgerchrome/blob/weapi/src/storage.js#L33) show domains with DNT set to false, though it looks like the setting will only ever be true or null. There doesn't appear to be a difference between using false or null so maybe either this or the comments above should be updated?

src/storage.js
```diff
@@ -0,0 +1,430 @@
...
+var setupUserAction = function(domain, action){
+  _setupDomainAction(domain, action, "userAction");
+};
+  
+/**
+* remove user set action from a domain
+* @param domain FQDN string
+**/
+var revertUserAction = function(domain){
+  _setupDomainAction(domain, null, "userAction");
```
rcutmore: Same as my comment above about reverting the DNT setting. Maybe either this should be updated to `_setupDomainAction(domain, "", "userAction");` or the [action_map comments](https://github.com/EFForg/privacybadgerchrome/blob/weapi/src/storage.js#L33) should be updated to show null for domains without a user override.

src/storage.js
cooperq: You are totally right! Completely unnecessary type casting going on here.

src/storage.js
cooperq: Yes! I don't know why I did this but a boolean is the correct choice here.

src/storage.js
cooperq: I think you are right, this value should always be a boolean T/F I will update the code.

cooperq:
Now that @SwartzCr and @Hainish have looked this over and given their approval I am going to merge. Please file any new comments as separate issues.

PR #820, opened 2016-04-30
Fix options page refresh while origins filtered

rcutmore:
This fixes item 6 mentioned in my comment in #819, namely the refresh of the options page causing an existing filter to be reset for the list of origins. Now an action like filtering the list of origins and setting a user override for an origin won't cause all origins to be displayed after the refresh.


cooperq:
Thanks, I remember you writing this before, but I guess somehow it never got merged? Anyway this is a huge UX improvement!

rcutmore:
I thought so too but it was probably for the Firefox version.

PR #821, opened 2016-05-01
Save booleans instead of strings for settings in new storage engine

rcutmore:
This fixes item 5 in my comment in #819.

Some of the settings were being saved in the new storage engine as strings instead of booleans ("true"/"false" instead of true/false) which was messing up boolean checks in other areas of the extension. Whether or not the user has seen the comic was still being saved directly in the extension's local storage instead of the new storage engine so this fixes that as well.


cooperq:
Thanks! This is, of course, the right way to do things.

PR #845, opened 2016-05-24
WIP Firefox web extensions compatibility fixes

### LINKED ISSUE #762: Privacy Badger should merge into a single codebase using WEAPI and removing ABP
It seems like there are two mostly independent codebases being developed. Is this the case?

If so, maybe this can change soon. Mozilla is planning on providing an extension API that will be compatible with chrome and opera called  [WebExtension](https://developer.mozilla.org/en-US/Add-ons/WebExtensions). It will be stable [later this year](https://wiki.mozilla.org/RapidRelease/Calendar).

I had some time to try this, so I downloaded the [ff developer edition](https://www.mozilla.org/en-US/firefox/developer/) to test this out, and I started walking through the ["porting from google chrome" guide](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Porting_from_Google_Chrome).

The first problem was that the "unlimitedstorage" permission is not supported in firefox yet. But hopefully we won't need more than the default amount of storage (5mb?) initially. There were a few changes needed to the manifest.json. Then I tried to load the extension to firefox. Basically it didn't work, the PB icon shows up and I can open the menu but it is not blocking anything.

I tried to take a screenshot of the menu, but for some reason the menu disappears when the screenshot happens.
pb-menu-in-ff

Here is the settings page
pb-settings-in-ff

I'm not sure what the next steps to take here would be.

The branch I'm working on is [here](https://github.com/cowlicks/privacybadgerchrome/tree/port-to-ff).

### LINKED ISSUE #778: Write code to migrate from legacy pb-firefox storage engine to new storage engine

### LINKED ISSUE #779: Begin testing on firefox with web extensions API and fix any bugs

cooperq:
Makes privacy badger compatible with firefox web extensions API. 
FIxes #779 
Fixes #778 
Fixes #762 
cc @rcutmore @SwartzCr @Hainish


cooperq:
This is still a work in progress! 
Todo list:
=======
* fix incognito mode
* come up with a plan for migrating data see [#1214790](https://bugzilla.mozilla.org/show_bug.cgi?id=1214790)
* beat it until we feel confident there are no more bugs
* publish to AMO

src/background.js
```diff
@@ -666,22 +666,24 @@ chrome.tabs.onReplaced.addListener(function(addedTabId, removedTabId){
 // Listening for Avira Autopilot remote control UI
 // The Scout browser needs a "emergency off" switch in case Privacy Badger breaks a page.
 // The Privacy Badger UI will removed from the URL bar into the menu to achieve a cleaner UI in the future.
-chrome.runtime.onMessageExternal.addListener(
```
cowlicks: I think this is supported? https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/onMessageExternal

cooperq: It's not! The documentation says it's not supported in firefox. I put a guard clause around it so it will still work in chrome (which is all we care about in this particular case.)

cowlicks:
What is the problem with incognito mode? PB seems to work in incognito mode on chrome, and in private browsing mode on firefox.

src/migrations.js
```diff
@@ -24,7 +21,7 @@ require.scopes.migrations = (function() {
 var exports = {};
 exports.Migrations= {
   changePrivacySettings: function() {
-    if (!chrome.extension.inIncognitoContext) {
+    if (!chrome.extension.inIncognitoContext && chrome.privacy ) {
```
cowlicks: `browser.extension.inIncongnitoContext` is broken, I made a bug report [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1276025).

cowlicks: The issue here is that firefox does not have a mode like Chrome's split. Background pages are never run with `inIncognitoContext == true`.

I'm having a hard time debugging this in firefox. I can't figure out how to get a console, or log messages in context where `inIncognitoContext == true`.

rcutmore:
I'm seeing a few display issues with Firefox nightly, a couple of which I haven't looked into too closely yet.

On the options page, scrolling through the list of trackers causes the tracking icons at the top to scroll off-screen. I submitted PR #848 to fix this.

Everything looks scaled down slightly in the popup, which is causing some issues with tooltips (two examples below). The tooltips are fine on the options page.

tooltips1

tooltips2

Also on the popup the sliders aren't displayed correctly (seen above).

lib/compat.js
```diff
@@ -15,276 +15,7 @@
  * along with Adblock Plus.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-//
```
cowlicks: The ABP copyright above this could probably be removed too. Does this commit need to be in the firefox compat pull request? It seems kind of orthogonal.

src/migrations.js
cowlicks: So even in private browsing mode in firefox, scripts in the popup are report `inIncognitoContext === false`. Also leaving this here for posterity https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/Extension/inIncognitoContext

cowlicks: Currently, I'm checking for `inIncognitoContext` in a content script, then sending this as a message to be added to `pb.tabData[tab.id].inIncognito`. This is working in Chrome, but not still not firefox :/

cowlicks: Actually, the method above was working. There were other issues with the popup display which made me think otherwise.

My proposed solution: Check for incognito mode in a content script on each tab load, add this to `pb.tabData[tab.id].inIncognito`, then we'll refer to this `inIncognito` property to check for ingcognito mode.

cowlicks: My ongoing work on this is here: https://github.com/cowlicks/privacybadgerchrome/tree/firefox-compat

I'm going to dive in to implementing this today.

cooperq: So the specific context where this is important for private browsing is here: https://github.com/EFForg/privacybadgerchrome/blob/master/src/storage.js#L406-L411
The migrations code is only neccessary as a guard clause around a migration that doesn't run in firefox anyway so we don't need to worry about it. 
I'm not sure that marking the tab as incognito matters in this case, because the storage engine doesn't know or care what tab it's storing data for.

But if split mode doesn't wokr that's a big issue and maybe a blocker for firefox. 
The other issue that I was seeing here was that the tabData structure doesn't seem to work in a private browsing window, somehow the tab ids that it has are different from teh tab ids that the private window thinks it has. This is probably the cause of the popup issue that you saw.

cowlicks: I check for `extension.inIncognitoContext` in a content script and use it to set `pb.inIncognito`. Then changed `_syncStorage`, the way you mention [here](https://github.com/cowlicks/privacybadgerchrome/blob/firefox-compat/src/storage.js#L406-L410). This works around the webextension bug.

But things are still broken, changes I make to PB in FF private mode, still effect PB in regular FF.

To make this work without "split" mode, would we just have to guard other places we use storage with checks to `inIncognitoContext`? Or is it a larger issue?

cowlicks: It looks like we will have to switch to `"incognito": "spanning"`  mode if we want to do this. FF will soon reject extensions which use  split mode.

https://bugzilla.mozilla.org/show_bug.cgi?id=1277686

rcutmore: We would have to rework some things to use spanning mode; however, unless I'm misinterpreting [this](https://developer.chrome.com/extensions/manifest/incognito#spanning) it sounds like we won't be able to make spanning mode work for PB?

cowlicks: @rcutmore I think that will just prevent us from opening the options pages in incognito mode. This is a regression, but I don't think it is important enough to block this. We'd still be able to change settings for the incognito pb in the popup.

cowlicks: I'm working on converting to spanning to fix incognito mode in #860

cooperq:
Incognito mode is now fixed thanks to a huge amount of work by @cowlicks !  Now we just need to focus on a migration strategy. It looks like mozilla is planning to go with hybrid extensions for this. To make this work we will need to write a jetpack api extension that wraps our WEAPI extension. The jetpack extension will be able to get a copy of the legacy extension data and migrate it to the new extension. Once that is written and tested we just have to wait for hybrid extensions to land in the stable branch and then we can release!

cowlicks:
@cooperq Do you intend to do the remaining work in this PR? It is getting so large and complex that keeping up to date with master will become untenable.

cooperq:
Merging this at the excellent advice of @cowlicks! Everything seems to be working and it's time to cut a new release!

Issue #781, opened 2016-03-21
Add list of multi-domain first parties


catrope:
At the Wikimedia Foundation, we run web sites on a number of different domains that talk to each other and are essentially all the same first party. Privacy Badger recognizes that `en.wikipedia.org` and `de.wikipedia.org` are the same party, but it does not recognize that domains like `commons.wikimedia.org` and `fr.wiktionary.org` are also the same party. This frequently leads Privacy Badger to block AJAX requests from e.g. `*.wikipedia.org` domains to `*.wiktionary.org` domains, which breaks [our unified login system](https://phabricator.wikimedia.org/T108505) and [our new cross-wiki notifications beta feature](https://phabricator.wikimedia.org/T129764).

Putting our domains on the yellow list doesn't help, because these features are closely tied to user identity so they need cookies for authentication to work. Another way we could work around this is by publishing a DNT policy file, but we can't, at least not right now, since we have our own privacy policy that is almost-but-not-quite DNT-compliant (see [relevant policy paragraph](https://wikimediafoundation.org/wiki/Privacy_policy#Our_Response_to_Do_Not_Track_.28DNT.29_signals) and [FAQ  entry](https://wikimediafoundation.org/wiki/Privacy_policy/FAQ#DNTFAQ)).

I've been thinking that other web sites probably have a similar issue. For example, a request from `gmail.com` to `*.google.com` should also be treated as a first-party request rather than a third-party request. For that reason I propose adding a list of groups of domains that all belong to the same party, and have Privacy Badger treat a cross-domain request within the same domain group as a first-party request (i.e. allow it by default).

michael-oneill:
The W3C Tracking Preference Expression document (TPE) defines a same-party <https://www.w3.org/TR/tracking-dnt/#rep.same-party>  (array of strings representing domains) property in the JSON Tracking Status Resource. It was designed for this very  purpose, so browsers plus their extensions i.e. user agents can differentiate between domains managed by the first-party data controller from those that are not, and act accordingly.

 

It is used on the medium.com site, attached is a screen shot which shows it.

 

The TPE has been a Candidate Recommendation since last August.

 

 

Mike

 

PS I know it’s annoying it has a dash in it, but it is there and works quite well.

catrope:
> It is used on the medium.com site, attached is a screen shot which shows it.

[...]

PR #847, opened 2016-05-27
WIP: checking multidomain first parties.

cowlicks:
This fleshes out some ideas in #781


cooperq:
Merging this since I think that this is a good way to do this. Needs to be integrated still of course.  Nice work! :+1:

alexristich:
Oh, the joys of working on a new project with yet-to-be-integrated code :) I went and read some older issues and stumbled across #781, which led me here. I ended up implementing something similar for #864, though using @cooperq's original idea, which is essentially based on the format used for `publicSuffixList.js` and the cookie block list.

@cowlicks good call on building a reference array! I forgot about this - one of those weird JavaScript quirks that comes in handy every once in a while. The only precaution would be to make sure nothing modifies the reference array, as it affects everything else, but this should be easy enough.

Also, I see a number of issues which relate back to #781 - maybe those should be given a different label so they're more clearly associated with #781. "Good volunteer task" isn't quite the right fit anymore, since a fix is already in the works for these :)

PR #848, opened 2016-05-28
Fix tracking icons (firefox-compat branch)

rcutmore:
This fixes the positioning of the tracking icons for Firefox. On the options page scrolling through the list of trackers was causing the icons to scroll off-screen.


cooperq:
Sweet! That was super annoying! Thanks!

PR #849, opened 2016-05-29
Use tab method supported by Firefox and Chrome

rcutmore:
This line was throwing an error because the *getSelected* method is unsupported in Firefox. Maybe this section of code should just be removed though? Maybe it serves some purpose but it looks like it just refreshes the tracker list again after the options page finishes loading which seems unnecessary.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -391,7 +391,7 @@ function syncSettings(origin, userAction) {
 }
 
 document.addEventListener('DOMContentLoaded', function() {
-  chrome.tabs.getSelected(null, function(/*tab*/) {
+  chrome.tabs.query({currentWindow: true}, function(/*tab*/) {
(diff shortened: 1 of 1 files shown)


cooperq:
It looks like you wrote this code originally! https://github.com/rcutmore/privacybadgerchrome/commit/b26a9016e57e7003282ca7db5f52c2b541ffe08f

But if you think it's no longer necessary I'm fine with removing it. The options page is definitely badly in need of some love and refactoring.

cooperq:
oops nevermind I guess you just removed whitespace, git blame is not always great. 
Anyway, I think it's worth playing around with it to see if it is no longer necessary.

rcutmore:
Sounds good, I'll take a closer look.

PR #854, opened 2016-06-02
Remove missing localized string on options page (firefox-compat branch)

rcutmore:
This string isn't defined in any of the localization files. Chrome doesn't render missing localized text while Firefox renders it as "???".

Only tangentially related, would it be worth consolidating the *Social Widget Settings* and *General Settings* tabs, similar to the current Firefox version? i.e. have the social widget setting listed as a checkbox under general settings.

### DIFF
--- a/skin/options.html
+++ b/skin/options.html
@@ -217,7 +217,6 @@ a, a:hover, a:active, a:visited{
   </div>
 
   <div id="tab-general-settings">
-    <p class="i18n_general_settings_help"></p>
     <form id="settingsForm" action="#">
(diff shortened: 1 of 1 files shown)


cooperq:
Thanks, good find! 
Yes I think we should absolutely move the social widget settings button onto the general settings tab as a checkbox. That would be great!

rcutmore:
Sounds good, I'll work on this. Maybe I'll wait until after some of the changes for Firefox are finished to avoid merge conflicts.

PR #856, opened 2016-06-03
Remove duplicate refresh while opening options (firefox-compat branch)

rcutmore:
Original discussion in #849.

This code does indeed cause an extra unnecessary refresh of the tracker list when opening the options page. What ends up happening is the tracker list refreshes twice in quick succession and the domains being displayed during both refreshes are exactly the same.

Removing this makes opening the options page a little snappier as well.


cooperq:
Thanks!!

PR #881, opened 2016-07-17
Enabling support for cross domain web properties which share assets

### LINKED ISSUE #864: Google currency charts currently blocked
Looking for currency conversions on Google, such as e.g. https://www.google.co.uk/search?q=1usd+to+gbp, in general produces a widget with a currency conversion tool and a chart of the recent history of the exchange rate. Since recently, however, Privacy Badger blocks this chart:

selection_034

This impairs the usability of a site that the user is explicitly asking for, and it does not expose the user to additional external tracking - the user is already visiting Google under their T&Cs, and the chart is provided by Google. Thus Privacy Badger should allow the chart through. (Whether this can be done without creating an ugly exception, of course, is another matter.)

alexristich:
This should fix #864; in addition to checking the base domain, it will also reference the cross domain list and see if the document and request share the same "owner".

Building the crossDomainList.js file is tedious, but no better option came to mind. The idea is that, for popular domains, the list can be added upon, which will hopefully ensure greater compatibility.

If this approach seems sound, there's one lingering question: should something like "youtube.com" be part of the "google" property as well?


cowlicks:
Hi @alexristich, I think this is generally a good approach. And is essentially the same as the idea I had but simpler.

It doesn't look like the test failure is related to your PR. You can restart Travis-CI by closing and reopening this PR.

But it would probably be good to have some tests for this. An integration test on pbtest.org would be nice, but that can't happen until pbtest is OSS (cc @cooperq).

lib/basedomain.js
```diff
@@ -203,8 +204,13 @@ function isThirdParty(/**String*/ requestHost, /**String*/ documentHost)
 
   // Extract domain name - leave IP addresses unchanged, otherwise leave only base domain
   var documentDomain = getBaseDomain(documentHost);
-  if (requestHost.length > documentDomain.length)
+  if (requestHost.length > documentDomain.length) {
+    var requestDomain = getBaseDomain(requestHost);
```
cowlicks: Maybe this should get wrapped up into an `isMultiDomainFirstParty` function.

alexristich:
Thanks for the feedback @cowlicks. I recognized after the fact that you'd already done this, but in a different place in the code (and with a better lookup algorithm, though as you say a little more complex!). Will spend some time familiarizing myself with the testing setup.

Just pushed an updated version of this - is this what you meant when you suggested it be wrapped in a helper?

cooperq:
@cowlicks @alexristich it sounds like we should probably go with cowlicks' implementation of the first party domains list, but the machinery of how that would be checked that you have made here looks good. Let's try to synthesize this into a different pull request? I will see if I can get to it today. If I do I will ask the two of you to review it if you don't mind :smile:

alexristich:
@cooperq Sure, that sounds great!

PR #899, opened 2016-08-24
Fix several errors.

cowlicks:
The commits explain them pretty well.

* Fix errors with `updateCount` for no tab requests. Internal xhrs were causing us to update pb's badge. But these don't have an associated tab, which was causing the error.

* Remove duplicate startup messages. The 'pb is ready to rock' message is kind of a lie now since the initialization funcs are asynchronous. But it is close enough.

* Fix error in migrations. This is bug was due to my refactoring. The `storage.getBadgerStorageObject` is a method on a badger instance now instead of a function. But that wasn't changed in this migrations code.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -67,8 +67,6 @@ function Badger(tabData, isIncognito) {
     // TODO: register all privacy badger listeners here in the storage callback
 
     badger.INITIALIZED = true;
-    console.log('privacy badger is ready to rock');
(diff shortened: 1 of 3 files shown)


cowlicks:
Also, I looked in to suppressing the 404 errors in the console that we get for requesting `https://somesite.com/.../dnt-policy.txt`but this is apparently impossible by design.

cooperq:
Awesome! Thanks!

PR #900, opened 2016-08-25
Chinese translation

paperbag:
And I added up "show_counter_checkbox".

ps.The former translation like machine translation.. So I changed some expressions of Chinese.
And should zh rename as zh_CN(because language code of simplified Chinese is zh_CN. )?


cooperq:
Thanks!

PR #910, opened 2016-09-01
Multi Domain First Parties

### LINKED ISSUE #781: Add list of multi-domain first parties

cooperq:
Fixes #781 adds awareness of domains which are techincally third party but actually belong to the same first party. 

cc @cowlicks @Hainish @alexristich


src/multiDomainFirstParties.js
```diff
@@ -11,15 +11,50 @@
...
+  ["wikia.com", "wikia.net", "nocookie.net"],
+  ["tripadvisor.com", "tacdn.com", "tamgrt.com"],
+  ["baidu.com", "bdimg.com", "bdstatic.com"],
+  ["wikipedia.org", "wikimedia.org", "wikimediafoundation.org", "wiktionary.org", 
+    "wikiquote.org", "wikibooks.org", "wikisource.org", "wikinews.org", 
+    "wikiversity.org", "mediawiki.org", "wikidata.org", "wikivoyage.org"],
+  ["twitter.com", "twimg.com", "t.co"],
+  [ "amazon.com", "amazon.ca", "amazon.co.jp", "amazon.co.uk", "amazon.de", 
+    "amazon.es", "amazon.fr", "amazon.it"],
+  ["pornhub.com", "phncdn.com"],
```
cowlicks: https://www.youtube.com/watch?v=3gOHvDP_vCs

cooperq: Just giving the people what they want ;)

alexristich:
Took a look at the code and it looks good. Is "dummy" there as a placeholder for additional properties to be added, or was that primarily for testing? 

For the array itself, is it better to have this within the code for the extension or host it on the EFF site in a JSON file like the DNT-compliance list and refresh it daily?

cooperq:
Dummy is there just so we don't have to remember to add or remove a comma at the end each time. IDK, it could be removed if it's too confusing. 
We could have the first parties array in a JSON file, it really depends how often we think we are going to have to update it. I suspect it won't have to get updated that often, but if that turns out to not be the case I think it should be pretty easy to refactor.

alexristich:
True, true. I suspect it might occur somewhat frequently judging by some of the recent tickets, but as you say, refactoring shouldn't be much of an issue down the line.

cowlicks:
LGTM. It'd be nice to have an integration test on pbtest.org.

PR #912, opened 2016-09-02
remove duplicate keys in JSON

X-dark:
This makes FF validation fail


cooperq:
Thanks! This is great!

PR #915, opened 2016-09-08
Consolidate settings

rcutmore:
This cleans up the settings page a bit by moving the social widget replacement setting into the general settings tab. I took a stab at updating all the localized strings as well but some may need to be fixed by others.


cooperq:
No idea why the test are failing on travis but they pass locally so I'm just gonna merge this and see what happens. Good work! Thanks!

Issue #919, opened 2016-09-12
"Report Broken Site" wording somewhat unclear


alexristich:
I've been talking to several people who are users of the extension, and many of them didn't understand the purpose of the "Report Broken Site" button. I think there is some room for improvement in the wording. The main source of confusion seems to stem from the ambiguity of what "broken" means in this case. Specifically, I offer up the suggestion, "Did Privacy Badger break this site? Let us know!"

Also, a stylistic suggestion - it would be nice if the font size was increased slightly for the text on these buttons.

cooperq:
Seems like a good issue to make a patch for :wink:

KidPluto:
Looks fixed, should be closed.  Thanks.

alexristich:
Whoops, thanks for pointing this out! Will close it now :)

PR #921, opened 2016-09-12
Rename 'Report Broken Site' button and adjust some font sizes

alexristich:
Tweaks in response to #919. I should note that my CSS is limited, so I don't know if the way I adjusted the font size for `<p>` and `<span>` is in any way best practice. Feel free to make adjustments and leave comments!

### DIFF
--- a/skin/popup.css
+++ b/skin/popup.css
@@ -377,3 +377,12 @@ font-size: 16px;
 .hidden {
   display: none;
 }
+#overlay p {
+  font-size: 12px;
+}
+#overlay label {
+  font-size: 13px;
+}
+#siteControls button {
+  font-size: 12px;
+}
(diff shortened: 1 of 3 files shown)


skin/popup.css
```diff
@@ -377,3 +377,12 @@ font-size: 16px;
...
   display: none;
 }
+p {
+  font-size: 12px;
+}
+span {
+  font-size: 12px;
+}
+#overlay label {
+  font-size: 13px;
```
cooperq: These are all too broadly scoped. You need to scope the font size change specifically to the buttons, this makes the text for the entire plugin very tiny.

alexristich:
I was worried about that. What I was specifically trying to fix, in addition to the size of the text on the buttons is the size of the text on the "Report Error" page itself, which is also microscopic.

pb-report-error-curr
pb-report-error-new

I've figured out a better way to do this without adding what are effectively globals. Let me know if this update looks better.

cooperq:
Looks great! Thanks!

Issue #591, opened 2015-08-29
Scrolling the list of trackers is nonobvious on mac computers


Zakkai:
![screen shot 2015-08-28 at 10 38 14 pm](https://cloud.githubusercontent.com/assets/1514445/9560190/d056653c-4dd5-11e5-86d5-9361e30c8d67.png)

On Google docs, where there are lots of trackers listed by the Badger, if mouse over the bottom listing, the pop-up tip is obscured under the "Disable Privacy Badger for This Site" button.

I'm in Chrome, running OSX Yosemite 10.10.1

cooperq:
You can scroll the list of domains down. then you should see the popup for the bottom listing.

Zakkai:
That's true, but in Chrome, the scroll bar doesn't display by default, until you begin to scroll. This means that there is no visual queue that scrolling is what you need to do, and I, a relatively savvy user, didn't guess it without the visual queue. I think it'd be great if there was some indication that scrolling was necessary. Perhaps a permanently visible scroll bar, or a border around the scrollable region.

Just my suggestion -- I know you have to prioritize more serious problems most of the time.

cooperq:
Yea that seems to be the case with OSX that they hide the scrollbar until you use it. Which is bad UI IMO. We should think about how to indicate to a user that they can scroll down.

karisafari:
I was initially having trouble replicating this issue--turns out I'd set my OS system preferences to always show scroll bars.

So if you'd like to fix your own user experience with the Privacy Badger plugin (and Chrome in general):
- Open "System Preferences"
- Select "General"
- Select the "Show Scroll Bars" -> "Always" radio button

That being said, I'll still have a look into fixing this for everyone else...

Zakkai:
Ahh, good to know.


Z.

SwartzCr:
this seems like a pretty big problem, there's no way we can override this in chrome or ff settings on macos?

Zakkai:
I feel like putting a border around the scrollable region would be good
enough, and wouldn't be confusing if you had scroll bars turned on.


Z.

cooperq:
we could look into a setting, but it seems unlikely so far since it is a part of the OS chrome. Maybe a border region would work, but I'm not convinced of that. I think we should figure out if there is a setting we can tweak first of all. If there is not, I think we should consider and do some testing about what might be the best way to indicate that there is a region below. I wonder if we could trick the scrollbar into appearing by jittering the window?

[...]

PR #927, opened 2016-09-14
Tidying up scrollbar in popup

alexristich:
Turns out the scrollbar *is* showing by default (re #591); not sure what I was looking at before. However, it was getting cut off. I made a minor tweak to the CSS to set the properties for the `clickerContainer` class rather than the `blockedResources` id. `blockedResources` contains `clickerContainer` but also contains the blocked, cookieblocked and allowed icons at the top. Because of this, the scrollbar was starting from the top of the box, whereas it should have started from beneath those icons.

Here are before and after screenshots:
scroll before
scroll after

### DIFF
--- a/skin/popup.css
+++ b/skin/popup.css
@@ -220,8 +220,8 @@ font-size: 16px;
 #noBlockingLink a {
   color:#4183c4;
 }
-#blockedResources{
-  max-height: 320px;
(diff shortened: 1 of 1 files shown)


cooperq:
awesome! Thanks!

PR #933, opened 2016-09-20
Run Travis tests for Chrome beta channel.

gunesacar:
Allow failures for the beta version.
Enable fast finishing.

Run tests with Python 3.4.
Fix https://github.com/EFForg/privacybadgerchrome/issues/937 by updating the print statements in the tests.

Don't create a virtual environment for the tests. Travis already creates one.
We can't seem to use the right Python (3) version if we create the virtual env. ourselves.

Remove redundant printing of the downloaded PB policy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,17 +4,33 @@ os: linux
 dist: trusty
 python:
   - "2.7"
+env:
+  matrix:
+    - TRAVIS_EXTRA_JOB_WORKAROUND=true
+matrix:
+  fast_finish: true
+  include:
+    - python: "2.7"
+      env: CHROME="google-chrome-stable"
+    - python: "2.7"
+      env: CHROME_PATH="google-chrome-beta"
+  allow_failures:
+    - python: "2.7"
+      env: CHROME_PATH="google-chrome-beta"
+  exclude:
+    - env: TRAVIS_EXTRA_JOB_WORKAROUND=true
 addons:
   apt:
     sources:
     - google-chrome
     packages:
     - google-chrome-stable
+    - google-chrome-beta
     - python-virtualenv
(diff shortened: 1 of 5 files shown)


gunesacar:
Run the Travis tests with the Chrome's beta channel. May help with foreseeing the problems with the upcoming releases.

This change shouldn't affect the existing tests:
* We use [`allow_failures`](https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail) to make sure the beta test failures will not affect the outcome of the stable tests.
* We use [`fast_finish: true`](https://docs.travis-ci.com/user/customizing-the-build#Fast-Finishing) to not wait for the beta tests to finish.

cooperq:
Is there a reason for us to run the tests in 2.7 and 3.4 ? The version of python shouldn't make any difference in the tests passing yes? Should we just pick one or the other?

cooperq:
This all looks good to me modulo my question. Merging anyway.

gunesacar:
> Is there a reason for us to run the tests in 2.7 and 3.4 ? The version of python shouldn't make any difference in the tests passing yes? Should we just pick one or the other?

This is to make sure we're not breaking the test suite for the contributors [who use Python 3 to run the tests locally](https://github.com/EFForg/privacybadgerchrome/issues/937#issue-178196154). 

You're sure right: Python version shouldn't affect the outcome of the tests. Rather, this is a _test for the test code_ itself.

There's no time overhead of running this extra test, but let's roll back if you think this is not a good idea. (I don't like the new `print` with the parenthesis anyways :smiling_imp:)

cooperq:
Makes sense to me! Let's leave it as is.

PR #938, opened 2016-09-22
Updating logic to make "user_allow" list be generated only on initialization

alexristich:
I modified the logic to only generate the `user_allow` list on launch. One improvement might be to clear out `user_allow` on init then recreate it, though there isn't currently a function for this within pb.storage, so I opted to leave it out for now.

I wrote some tests to make sure `user_allow`is being updated properly when changes are initiated by a user. Feedback is welcome.

(Also, two add'l commits are appearing because I merged master into my branch, rather than merging my branch into master then running tests. Not sure if these will appear in the commit history; should I rebase?)


src/background.js
```diff
@@ -238,6 +239,22 @@ Badger.prototype = {
   },
 
   /**
+   * Search through action_map list and update list of domains
+   * that user has manually set to "allow"
+   */
+  updateUserAllowList: function() {
+    var self = this;
```
rcutmore: It doesn't look like this is needed.

src/webrequest.js
```diff
@@ -584,13 +584,13 @@ function getSocialWidgetBlockList(tabId) {
...
   var badger = getBadgerWithTab(tabId);
-  var green_domains = badger.storage.getAllDomainsByPresumedAction(constants.USER_ALLOW);
+  var user_allow = badger.storage.getBadgerStorageObject('user_allow');
 
   window.SocialWidgetList.forEach(function(socialwidget) {
     var socialWidgetName = socialwidget.name;
 
     // replace them if the user hasn't greened them
-    if (green_domains.indexOf(socialwidget.domain) > -1 ) {
+    if (user_allow.hasItem(socialwidget.domain)) {
```
rcutmore: This section could be condensed:

    socialWidgetsToReplace[socialWidgetName] = !user_allow.hasItem(socialwidget.domain);

alexristich:
@rcutmore @cooperq Thanks for the feedback! I've moved the list into memory rather than onto disk. Also tidied up the code a bit. Let me know if there are any other issues!

src/webrequest.js
```diff
@@ -584,17 +584,12 @@ function getSocialWidgetBlockList(tabId) {
...
   window.SocialWidgetList.forEach(function(socialwidget) {
     var socialWidgetName = socialwidget.name;
 
-    // replace them if the user hasn't greened them
-    if (user_allow.hasItem(socialwidget.domain)) {
-      socialWidgetsToReplace[socialWidgetName] = false;
-    } else {
-      socialWidgetsToReplace[socialWidgetName] = true;
-    }
+    // replace them if the user hasn't manually allowed them
```
cooperq: Capitalize Replace

tests/.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
cooperq: remove and add to gitignore

src/storage.js
```diff
@@ -240,20 +240,34 @@ BadgerPen.prototype = {
...
+      pb.userAllow.push(domain);
+    }
+
     this._setupDomainAction(domain, action, "userAction");
   },
 
   /**
-  * remove user set action from a domain
+  * remove user set action from a domain and remove it from userAllow
+   * list in case it was previously allowed by user
```
cooperq: weird indentation please fix.

src/webrequest.js
cooperq: clarify comment

PR #941, opened 2016-09-25
Block ping requests sent by navigator.sendBeacon (issue #587).

### LINKED ISSUE #587: PrivacyBadger is missing blocking of HTML5 cookieless tracking
With HTML5 came a *new attribute to allow tracking of which links on a page are followed*, which means we don't even have to click them. The attribute in question is called the ***'ping' attribute***, and I don't see anything anywhere that addresses its abilities. PrivacyBadger should strip these attributes from each and every page to stop the use before its use becomes widespread.

gunesacar:
Until the following Chrome bugs are fixed and landed
PB can't know which tab or frame is the initiator of the ping
request. So PB will block all ping requests, even when it's
disabled on the site that sent the ping:

- https://crbug.com/522124 (landed in the Chrome unstable)
- https://crbug.com/522129

Once these bugs are landed, PB will treat the ping requests
as any other request and will block only when it's enabled for the
current tab.

This is tested with the Chrome unstable, where the fix
for bug-522124 is landed: since the tabId and frameId for the ping
requests are available, PB honors the disabled status and uses the
same blocking logic as the other requests.

Tested manually with Wireshark and debug logging from the extension.
Fixes #587

### DIFF
--- a/src/webrequest.js
+++ b/src/webrequest.js
@@ -55,6 +55,13 @@ function onBeforeRequest(details){
     recordFrame(details.tabId, details.frameId, details.parentFrameId, details.url);
   }
 
+  // Block ping requests sent by navigator.sendBeacon (see, #587)
+  // tabId for pings are always -1 due to Chrome bugs #522124 and #522129
+  // Once these bugs are fixed, PB will treat pings as any other request
+  if (type == "ping" && details.tabId < 0 ){
+    return {cancel: true};
+  }
+
   if ( _isTabChromeInternal(details.tabId)){
     return {};
   }


src/webrequest.js
```diff
@@ -55,6 +55,10 @@ function onBeforeRequest(details){
     recordFrame(details.tabId, details.frameId, details.parentFrameId, details.url);
   }
 
+  if (type == "ping" && details.tabId < 0 ){
```
cooperq: Are there ever ping requests that happen on visible tabs? Is there a reason to make this check? 
Also it might be having a comment which points to an explanation of what a ping request is, since this may be non obvious.

gunesacar: > Are there ever ping requests that happen on visible tabs? Is there a reason to make this check?

Yes, `navigator.sendBeacon()` can be called anytime, not only after the `onunload` event. I've seen Bing, Google Search, Google Analytics, Youtube and Moatads sending pings on visible tabs.

In the current Chrome stable,  tabId for visible tabs are missing due to https://crbug.com/522124. But this is already fixed and will land on release soon.

> Also it might be having a comment which points to an explanation of what a ping request is, since this may be non obvious.

Sure, I will add the comments. Thanks for the feedback!

cooperq: So if that is the case should we remove `details.tabId < 0` from the `if` clause?

gunesacar: If we want to blanket-block all pings, yes, we should remove the `tabID` check.

Very soon, we'll have access to tabId for the pings from visible tabs (#522124).
The `if` clause will make sure we treat pings as any other HTTP request (i.e. not blanket-block but check the origin, dest, blacklist etc.)

Let me know which one you'd prefer.

cooperq: I see, yes that makes sense. Thank you!

cooperq:
Going to go ahead and merge this. Thanks Gunes!

PR #947, opened 2016-10-02
Fix options page tracker list

rcutmore:
This fixes an issue with the options page caused by a duplicate div.

**Before**:
screenshot at 2016-10-02 16 05 53

**After**:
screenshot at 2016-10-02 16 04 24


cooperq:
Looks great to me! Thanks!

PR #948, opened 2016-10-03
Adds @razorpay to the yellowlist.

captn3m0:
- We have faced issues with privacy badger blocking razorpay, and users being unable to pay before: https://github.com/hasgeek/boxoffice/issues/39

Yellowlisting would allow our users to pay easily. (See <https://razorpay.com/demo/> for a demo.)


cooperq:
Thanks!

captn3m0:
Thanks a lot :100:

PR #959, opened 2016-10-09
Open popup trackers link in new window

rcutmore:
Currently the "trackers" link on the popup doesn't work. This fixes that by changing the link to open in a new window.


cooperq:
Looks great! Thanks!

Issue #878, opened 2016-07-08
Add tests for popup to selenium tests.


cowlicks:

cowlicks:
@cooperq it doesn't look like this has been fixed. Is there a reason why we shouldn't add selenium tests for the popup?

cooperq:
Nope! I erroniously closed it thinking it was part of the incognito mode ticket , my mistake!

cowlicks:
Closed by #964

PR #964, opened 2016-10-12
Add tests for popup

rcutmore:
This addresses #878 but it has some shortcomings. From what I can tell it's not possible to open the popup normally with Selenium because Selenium can only interact with a webpage and not the other browser elements.

This PR works around that by opening the popup page directly and testing some of its functionality. Unfortunately this means the most critical part of the popup, the tracker list, isn't being tested. Maybe there's a way around that?

Before this can be merged, either I need to rebase after [PR 959](https://github.com/EFForg/privacybadgerchrome/pull/959) is merged or I need to remove the check for opening the tracker link in a new window.


cooperq:
Looks like github / travis was able to handle the rebase for you. Once I merged that PR and reran the tests they pass. This looks good to me modulo the issues you mentioned so I'm gonna go ahead and merge. Thanks for the commit!

PR #988, opened 2016-11-04
Toggle to enable strictest WebRTC IP leak protection

alexristich:
For #969:
I built off the existing branch; added a button to the settings and updated the logic accordingly. When Privacy Badger is enabled, it retains control of the in-browser WebRTC setting. When disabled, it returns control back to another extension or the browser itself.

RE: Firefox users, should I add logic in `loadOptions()` to set the visibility of this checkbox/setting to `hidden` if `chrome.privacy` is falsy?


skin/options.html
```diff
@@ -222,6 +230,21 @@
           <input type="checkbox" id="replace_social_widgets_checkbox">
           <span class="i18n_options_social_widgets_checkbox"></span>
         </label>
+      </div>
+      <div class="checkbox">
+        <label>
+          <input type="checkbox" id="toggle_webrtc_mode">
+          <!--TODO add string to i18n and update wording-->
+          <span>Prevent WebRTC from leaking local IP address*</span>
```
cooperq: internationalize these strings

alexristich:
This should now be ready to go.

alexristich:
Seems to be a _minor_ bug, but unrelated to the code changed here. If I manually change Chrome's internal `WebRTCIPHandlingPolicy` setting from the console and refresh the options page once, our setting doesn't update. When I refresh it a second time, it does update. I did this a couple times locally and it appeared reproducible. Not sure if this is some caching optimization, but ideally the `General Settings` tab should refresh whenever a user navigates to it to make sure it's up to date (and doesn't require refreshing the options page). Still, very much an edge case that can probably be picked up in 2.x.

PR #1006, opened 2016-11-30
Note the support for wildcards in domain whitelists

ralish:
Before now this was only known by reading the code of `src/background.js` or the single reference in the [Changelog](doc/Changelog) for the `v1.0.4` release. This will also require updating translations but I'm not sure what the process for that is?

### DIFF
--- a/skin/options.html
+++ b/skin/options.html
@@ -172,7 +172,7 @@ a, a:hover, a:active, a:visited{
       <table>
         <tr>
           <td style="max-width:100%">
-            <input type="text" value="" id="newWhitelistDomain" style="width:100%">
+            <input type="text" value="" id="newWhitelistDomain" style="width:100%" placeholder="e.g. www.domain.com, *.domain.net, domain.org">
           </td>
           <td>
             <button class="addButton" type="submit"><span class="i18n_add_domain_button"></span></button>


alexristich:
Very good point. Perhaps putting some grayed-out text like "_e.g. www.sample.com, *.sample.com, etc._ in the input field might be clearer yet. This would also potentially avoid the trouble of editing the translation files :)

On the translations, here's one of my recent PRs where I had to add to the translation files: https://github.com/EFForg/privacybadgerchrome/pull/982/files

When adding a new translation, the steps are basically (1) adding string to the `messages.json` file for all languages and (2) using the `i18n_{{string name}}` class in HTML or using the `i18n.getMessage()` function from `chrome.i18n` in JS (depending on where the string appears).

ralish:
@alexristich Yeah, that's a great point, if we can avoid updating the translation files for now, that would be a win. I've updated the PR to add placeholder text to the input field as you suggested. I've made the change directly in `skin/options.html` on the grounds this is not an i18n string, or at least, not until we want to have different sample domains based on the user's language. That can probably be a future change though! Thoughts?

alexristich:
Excellent, will take a closer look tomorrow to make sure this looks good!

alexristich:
Adds placeholder as per #1017.

Issue #1008, opened 2016-11-30
"Show count of blocked items" unchecked by default


cooperq:
The "show count of blocked items" setting (under general settings) is unchecked by default even though it is enabled by default. We should fix this. 

This might be a good task for @zebesta

cooperq:
fixed by @zebesta thanks!

PR #1020, opened 2016-12-02
Fixes issue #1008

zebesta:
This fixes the checkbox issue as the HTML was improperly named and disconnected from the javascript code. #1008


cooperq:
Awesome, thank you so much!

PR #1037, opened 2016-12-08
Removing download permission and adding workaround from uBlock

alexristich:
This reverts the permission that was added in `81c8fbb3fcc5a48e726c91d175587b55beddac42` and adds a workaround from uBlock that allows downloading of a file without going through `chrome.downloads`.


cooperq:
lgtm

PR #1040, opened 2016-12-10
add release utils to repo

cooperq:
adds release utils which were previously in a separate repo for chrome. This is a merging of the already existant chrome and firefox release utils.


cooperq:
Merging TBR @ghostwords

PR #1061, opened 2016-12-14
fix #1048: dynamic width & calc() width of instruction

### LINKED ISSUE #1048: Version 2016.12.8’s panel gets cut off in Firefox menu
I placed the privacy badger amongst the icons in my Firefox menu on the top right. When I click it there the other icons slide to left and a panel opens. Since the update today this panel does no longer fit into the given space but is cut on the right side. It’s not possible to operate the sliders then.

pb-fixed-width

The earlier version had a flexible width and adapted to its container.

BTW: The new panel also seems to be very confusing and untidy. I think the font size of the domains and the buttons are now smaller, the domains have a grey background, buttons have become frames and there’s a misplaced white space under each domain – probably to reserve space for the almost information-free tool tips. I’m a little bit disappointed.

groovecoder:


cooperq:
lgtm

PR #1091, opened 2016-12-17
Update messages.json

Urides:
Forgot a little typo, sorry.


Urides:
Entries from the original file in english are missing in the french version, i'll send a pull request with those when added.

PR #1097, opened 2016-12-18
Use some of the RTC components early on to see if they work

### LINKED ISSUE #1047: Privacy Badger has crashed Click this balloon to reload the extension
Privacy Badger Version 2016.12.8

Browser = http://chromium.woolyss.com/
Nik wpo cfg all-codecs stable 64-bit
No sync • No WebRTC • No Widevine • 55.0.2883.87 (423768) • Friday, 9 Dec 2016

Problem = Every time I click the Baloon to reload Privacy Badger it crashes again and again and again

I have uninstalled it in case of file corruption, then re-installed the latest version again .. No joy the same problem immediately starts again.

qbit:
Chrome seems to expose the WebRTC api even if it is disabled at compile time.
This extra check attempts to create a peer connection to see if the components
actually work.

This resolves #1047 for me, but I don't have a good setup to test
other (webrtc enabled) boxes.

This check is similar to what [uBlock Origin](https://github.com/gorhill/uBlock/blob/master/platform/chromium/is-webrtc-supported.js) does.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -179,6 +179,19 @@ Badger.prototype = {
     if (!chrome.privacy) {return;}
     var cpn = chrome.privacy.network;
 
+    // Chrome still has various RTC mechanisms when webRTC is disabled
+    // at build time. Check to see if we can actually use some of them.
+    try {
+      var RTCPC = window.webkitRTCPeerConnection;
+      var pc = new RTCPC(null);
+    } catch(err) {
+      return;
+    }
+
+    if (pc != null && pc.close) {
+      pc.close();
+    }
+
     var settings = this.storage.getBadgerStorageObject("settings_map");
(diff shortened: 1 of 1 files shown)


alexristich:
Thanks for the pull request! Looks like it it failed on the lint check - would you take a look into that?

Adding this also probably requires some additional tweaking, such as storing the result as a property of the `badger` object. This setting could then determine whether the WebRTC option should show up on the Options page.

I want to let @cooperq give this a look before merging. It might take a little bit, since things slow down around the holidays. However, if you can fix the linting errors it shouldn't be too contentious :)

qbit:
Fixed the linting issues!

cooperq:
Just got around to looking at this. It looks like #1108 is the same but a bit more robust so I am going to accept that one instead but thank you for your contribution!

Issue #1063, opened 2016-12-14
Can't export data using Opera


redranamber:
From the console

Error in response to storage.get: InvalidCharacterError: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
    at Object.callback (chrome-extension://ldfkcgjipgfchpnojicdgpgiocoeelik/src/options.js:169:57)
    at HTMLButtonElement.exportUserData (chrome-extension://ldfkcgjipgfchpnojicdgpgiocoeelik/src/options.js:166:24)
    at HTMLButtonElement.dispatch (chrome-extension://ldfkcgjipgfchpnojicdgpgiocoeelik/lib/jquery-ui/js/jquery-2.2.4.min.js:3:7537)
    at HTMLButtonElement.r.handle (chrome-extension://ldfkcgjipgfchpnojicdgpgiocoeelik/lib/jquery-ui/js/jquery-2.2.4.min.js:3:5620)

alexristich:
Thanks for the report on this!

@cooperq @ghostwords I notice uBlock is using a different approach for creating the export file. 
```
    var filename = vAPI.i18n('whitelistExportFilename')
        .replace('{{datetime}}', uBlockDashboard.dateNowToSensibleString())
        .replace(/ +/g, '_');
    vAPI.download({
        'url': 'data:text/plain;charset=utf-8,' + encodeURIComponent(val + '\n'),
        'filename': filename
    });
```

Our current implementation is using `data:application/json;base64` instead - I'd assumed base-64 would give us the character set we needed though it seems not. Should I switch `btoa()` to `encodeURIComponent()` along with the required tweaks to the file format?

cooperq:
sounds good @alexristich I will take a look at your PR

cooperq:
fixed by #1109

PR #1109, opened 2016-12-23
Changing export to use UTF-8 instead of Base64

alexristich:
@ghostwords 

As per #1063, Base64 doesn't support a wide enough character set to handle all possible URLs (e.g. URLs can now be in non-Latin alphabets). This changes the generated download URL to be in UTF-8 encoding and uses the relevant conversion function to support this set of characters.


cooperq:
lgtm

PR #1129, opened 2017-01-09
Create messages.json

Sijera:
Serbian


cooperq:
I fixed one typo (double quotes needed escaping.) Looks great other than that! Thanks for the great work!

PR #1147, opened 2017-01-20
Update italian translation

anthologist:
Also, why all the translated slideshow strings are not used? Is it a bug?


cooperq:
Thank you very much!

Issue #1052, opened 2016-12-13
suggestions for src/multiDomainFirstParties.js 


jawz101:
Taken from Lastpass Equivalent Domains and formatted for PrivacyBadger, fyi

["ameritrade.com","tdameritrade.com"]
["bankofamerica.com","bofa.com","mbna.com","usecfo.com"]
["sprint.com","sprintpcs.com","nextel.com"]
["youtube.com","google.com","gmail.com"]
["apple.com","icloud.com"]
["wellsfargo.com","wf.com"]
["southerncompany.com","southernco.com"]
["mymerrill.com","ml.com","merrilledge.com"]
["accountonline.com","citi.com","citibank.com","citicards.com","citibankonline.com"]
["cnet.com","cnettv.com","com.com","download.com","news.com","search.com","upload.com"]
["bananarepublic.com","gap.com","oldnavy.com","piperlime.com"]
["bing.com","hotmail.com","live.com","microsoft.com","msn.com","passport.net","windows.com"]
["ua2go.com","ual.com","united.com","unitedwifi.com"]
["overture.com","yahoo.com","flickr.com"]
["zonealarm.com","zonelabs.com"]
["paypal.com","paypal-search.com"]
["avon.com","youravon.com"]
["diapers.com","soap.com","wag.com","yoyo.com","beautybar.com","casa.com","afterschool.com","vine.com","bookworm.com","look.com","vinemarket.com"]
["1800contacts.com","800contacts.com"]
["amazon.com","amazon.co.uk","amazon.ca","amazon.de","amazon.fr","amazon.es","amazon.it","amazon.com.au"]
["express-scripts.com","medcohealth.com"]
["mynortonaccount.com","norton.com"]
["verizon.com","verizon.net"]
["logmein.com","logme.in"]
["rakuten.com","buy.com"]
["siriusxm.com","sirius.com"]
["ea.com","origin.com","play4free.com","tiberiumalliance.com"]
["37signals.com","basecamp.com","basecamphq.com","highrisehq.com"]
["steampowered.com","steamcommunity.com"]
["chart.io","chartio.com"]
["gotomeeting.com","citrixonline.com"]
["gogoair.com","gogoinflight.com"]
["mysql.com","oracle.com"]
["discover.com","discovercard.com"]
["dcu.org","dcu-online.org"]
["nefcuonline.com","nefcu.com"]
["pepco.com","pepcoholdings.com"]
["century21.com","21online.com"]
[...]

PR #1173, opened 2017-01-30
Update multiDomainFirstParties.js

jawz101:
based on single sign-on equivalent domains https://github.com/EFForg/privacybadger/issues/1052


jawz101:
I don't think the failed check is from my commit.

cooperq:
LGTM Thanks very much!

PR #1193, opened 2017-02-08
Create messages.json

Clonewayx:
Translate to Czech

### DIFF
--- a/_locales/cs/messages.json
+++ b/_locales/cs/messages.json
@@ -1,0 +1,300 @@
+{
+    "add_domain_button": {
+        "message": "Přidat doménu"
+    },
+    "badger_status_block": {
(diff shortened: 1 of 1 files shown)


cowlicks:
Thank you! I *think* that translations are handled on the [transifex site](https://www.transifex.com/eff/privacy-badger/). So the maintainers might add these there, instead of merging here.

cowlicks:
Oh! I looked at previous PR's, and it seems translations *can* be merged here. Sorry for the noise.

cooperq:
Yes, to clarify we don't actually do any translations on transifex (mostly because I have never taken the time to figure out how it works) so this is fine! Thanks!

cooperq:
This looks great! Thanks for the translation!

PR #1199, opened 2017-02-13
Remove superfluous rsa.js and jsbn.js dependencies. 

cowlicks:
These might have been inherited with the old Addblock Plus code. They don't seem to be used anywhere in the codebase.


cooperq:
Thanks! Yea I think these were ABP artifacts.

PR #1168, opened 2017-01-28
Selenium tests in Firefox

cowlicks:
Currently the support for FireFoxe's WebExtensions in selenium is lacking. As of today there is no way to install a WebExtension   [#1181](https://github.com/seleniumhq/selenium/issues/1181) with selenium, or geckodriver [#211](https://github.com/mozilla/geckodriver/issues/211). Fortunately there *is* a way to install them using [`web-ext`](https://github.com/mozilla/web-ext).

The strategy in this PR is to get selenium working for us:
- [x] install the extension and start firefox with `web-ext`
- [x] attach `geckodriver` to the running instance of firefox
- [x] connect `selenium.webdirver.Remote` to `geckodriver`
- [x] refactor the selenium tests to be browser agnostic
- [x] get this working with `xvfb` so we can run it on travis
- [x] run on travis

closes #1009  🎉 

I was experimenting with `xvfb` earlier today, so that step might not be too hard.

Have y'all seen anyone else who has this working yet? This solution is kind of a hack, but it is only temporary until selenium gets up to speed.

PR #1205, opened 2017-02-14
Fix broken selenium cookie tracker tests.

cowlicks:
This fixes `tests/selenium/cookie_test.py::CookieTest.test_cookie_tracker_detection` and cleans up the test file. In theory should work on top of #1168, but should probably be re-tested with it when it is merged, or vice-versa.

The main issue seemed to be the test was not python3 compatible.


cowlicks:
Rebased.

cooperq:
Looks great! Thanks!

PR #1209, opened 2017-02-16
Remove `lib/sha1.js` dependency

### LINKED ISSUE #1200: Remove `lib/sha1.js` dependency and use the builtin webcrypto API
We can replace the sha1 function ([used here](https://github.com/EFForg/privacybadger/blob/master/src/background.js#L368)) with the builtin webcrypto api. It should look something like [this](https://github.com/cowlicks/bypasser/blob/master/src/utils.js#L38-L46), but account for whatever format the response is in.

```javascript 
function doHash(bytes) {
  return window.crypto.subtle.digest(
    { name: "SHA-1", },
    Uint8Array.from(bytes)
  )
  .then(hashed => {
    return new Uint8Array(hashed);
  });
}
```

cowlicks:
closes #1200


cowlicks:
rebased

cooperq:
This mostly looks good. I think it might be worth having a seperate test for sha1 as well. I would like to see that before merging.

PR #1212, opened 2017-02-18
rename sample_cookieblock* to cookieblock*

cowlicks:
The file `doc/sample_cookieblocklist.txt` is the canonical  cookieblocklist, so having it prefixed with `sample_` is confusing. It should also be removed from the `doc/` directory since it isn't documentation. Where would be a better place? `data/`?


cooperq:
You are technically correct, however this also requires us to maintain 4 seperate cookie block lists until everyone has updated to the new cookie block list location. I'm not excited about doing this right now.

PR #1216, opened 2017-02-20
Reverting change from 4124c86 to fix issue with toggle switch (#1197)

alexristich:
As per #1197, the sliders were broken quite a while back. I tracked this down finally to an issue with the extraction of the anonymous function in `$('.switch-toggle').each(...)` into an externally defined function. It isn't exactly clear to me what in this change actually affected the slider, so it's not at all surprising that this got overlooked. Maybe a closer look into this will yield some further information, though until then this simple reversion seems to fix the issue.

Here's an updated image to compare with that from the original issue:
slider-fixed

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -421,23 +421,23 @@ function showTrackingDomains(domains) {
     var value = $(this).children('input:checked').val();
 
     var slider = $('<div></div>').slider({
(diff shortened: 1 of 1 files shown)


cooperq:
fixes #1197

Issue #1113, opened 2016-12-25
chrome dropdown not showing full window


manarchist8691:
there's no way to scroll to the options at the bottom on this site. at least some other sites work ok.
![screenshot 2016-12-25 12 49 19](https://cloud.githubusercontent.com/assets/24766112/21472319/db17c9a8-caa1-11e6-99ef-932d0f392996.png)

cooperq:
Do you happen to have page zoom set to something other than 100% in chrome settings? That seems to cause this issue.

manarchist8691:
Yes, I default to 125%. Guess I can change that if need be. Thanks.

cooperq:
That would be a good workaround for now. We should probably support larger font sizes long term though.

lemnis:
This should not happen anymore since #1516, it should now add a scrollbar to the popup. @manarchist8691 , can you confirm this?

realnc:
It now has scrollbars for me.

Issue #1077, opened 2016-12-15
Badger popup UI is cut off on the bottom in some locales


pvsomebody:
Hi. The issue can be considered aesthetic now, since the button can still be clicked, and is now clipped only a little, and only on the domains that have a lot of trackers, but you asked to repost it if it happens after the update, so here.

![screenshot2](https://cloud.githubusercontent.com/assets/13513395/21222649/a02d3d88-c2cb-11e6-898d-25577dbfc4c7.png)

cooperq:
Thank you for bringing this to our attention!

airone506:
This issue does not seem to be limited to RU localization. See attached screenshot please.

Using Firefox 50.1.0 on Ubuntu. The Privacy Badger version is 2016.12.15.1.

![pb-01](https://cloud.githubusercontent.com/assets/22847930/21295505/cd69a394-c556-11e6-830a-62236728f0f8.png)

Urides:
Hi,

It's even worse in the french version of PB. The donate text doesn't even appear when there are too much trackers listed (PB 2017.3.28 with Firefox 52.0.2 on Linux Mint).

![uipb](https://cloud.githubusercontent.com/assets/24555008/25285912/37f08a72-26bc-11e7-8ac7-d0c859a8e95e.png)

Thanks for reading.

boxer4:
I believe I'm seeing the same thing in the en/US locale too after auto downloading the 5/9/2017 update.  The popup window size does not increase beyond the size of the three buttons normally seen for me, so the buttons and everything else beyond does not show up.  Bug #1252 has examples of what I see in en/US except the location is correct (like the ones in these other locales).  I reverted back to the 3/22/2017 as I had a backup of this version which restored proper operation.  Firefox 45.8.0/Linux.

MTco:
En/US here.

![2017-06-05_privacy-badger_2017 5 9](https://cloud.githubusercontent.com/assets/8866895/26786607/0cb3f326-49bc-11e7-81e4-f16efffe73e6.png)

ghostwords:
@MTco Which version of Firefox do you have?

@MTco @boxer4 Does the problem go away if you upgrade Firefox?

boxer4:
New Firefox ESR recently available from OS packager, PB 5.9.2017 seems to work properly with Firefox 52.1. Assuming there was an api change in Firefox between the ESRs.

ghostwords:
[...]

PR #1356, opened 2017-04-29
Adds a version number as per #1088

### LINKED ISSUE #1088: Suggestion - Show Version of Privacy Badger in dropdown
With the update to Privacy Badger 2.0 I am getting confused what machines I have updated and what ones I have not. I think a good enhancement would be to include text indicating the version or build in the text included in the drop down area. Perhaps right after the title "Privacy Badger" I have included a marked up screen shot.

This request is an enhancement in my opinion but figure having an easy way to tell what version is being used is always good, especially for troubleshooting purposes.

Thanks

Bradford

privacybadgerversionindicator

ynasser:
Fixes #1088.

I haven't really written any frontend code before, so I hope this isn't too bad.

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -94,5 +94,4 @@
     "skin/*",
     "icons/*"
   ],
-  "display_version": "2.0.0"
 }
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -93,5 +93,5 @@
   "web_accessible_resources": [
     "skin/*",
     "icons/*"
-  ],
+  ]
 }
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -93,5 +93,6 @@
   "web_accessible_resources": [
     "skin/*",
     "icons/*"
-  ]
+  ],
+  "display_version": "2.0.0"
 }
--- a/src/skin/popup.html
+++ b/src/skin/popup.html
@@ -66,6 +66,7 @@
   <a id="help" title="help" href='/skin/firstRun.html' class="control-button grey-button" style="display:block" target="_blank"><img src='/icons/help.svg'></a>
   <div id='title'><img src='/icons/badger-48.png'> <h2><span class="i18n_privacy_badger"></span></h2></div>
   <div class='clear'></div>
+  <div id="subtitle"><span id="version" class="i18n_version"></span></div>
 </div>
 <div id="blockedResourcesContainer">
   <p id="pbInstructions"> <span class="i18n_pb_detected"></span> <span id='number_trackers'></span> <span class="i18n_popup_instructions"></span></p>
@@ -80,6 +81,5 @@
   <center><button id="donate" class="pbButton" style="display:block" target="blank"><span class="i18n_donate_to_eff"></span></button></center>
 </div>
 <div class='clear'></div>
-<div class='footer'><span id="version" class="i18n_version"></span></div>
 </body>
 </html>
--- a/src/skin/popup.html
+++ b/src/skin/popup.html
@@ -64,7 +64,7 @@
 <div id='privacyBadgerHeader'>
   <a id="options" title="options" href='/skin/options.html' class="control-button grey-button" style="display:block" target="_blank"><img src='/icons/options.svg'></a>
   <a id="help" title="help" href='/skin/firstRun.html' class="control-button grey-button" style="display:block" target="_blank"><img src='/icons/help.svg'></a>
-    <div id='title'><img src='/icons/badger-48.png'> <h2><span class="i18n_privacy_badger"></span></h2></div>
+  <div id='title'><img src='/icons/badger-48.png'> <h2><span class="i18n_privacy_badger"></span></h2> <h2><span id="version_number"></span></h2></div>
   <div class='clear'></div>
 </div>
 <div id="blockedResourcesContainer">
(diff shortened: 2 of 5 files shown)


ynasser:
Thanks for the feedback. I've addressed some of the changes you requested, but couldn't figure out how to get localization to work. I updated `_locales/en_US/messages.json` to include `version` and tried using `i18n.getMessage("version")` in `popup.js` to no avail. Is there something more required to get localization to work?

src/manifest.json
```diff
@@ -93,5 +93,5 @@
   "web_accessible_resources": [
     "skin/*",
     "icons/*"
-  ]
+  ],
```
ghostwords: The trailing comma makes it invalid JSON (JSON is more constrained than plain JavaScript). Chrome doesn't let you load the extension; all the tests fail because of this.

src/skin/popup.html
```diff
@@ -80,5 +80,6 @@ <h1 id="report_title" class="i18n_report_title"></h1>
   <center><button id="donate" class="pbButton" style="display:block" target="blank"><span class="i18n_donate_to_eff"></span></button></center>
 </div>
 <div class='clear'></div>
+<div class='footer'><span id="version" class="i18n_version"></span></div>
```
ghostwords: I like this more, but I have another suggestion: Could we try positioning the version string directly beneath the words Privacy Badger? Mockup attached below:

screenshot from 2017-05-01 11 39 49

Here are the suggested changes:
- Reduce size of "Privacy Badger"
- Move "Privacy Badger" up a little bit to make room for the version string below
- Align "Privacy Badger" and version string along left edge
- Create white space below the title to give some breathing room to the title and the summary text below the title

ghostwords: I think the localization is working (it says "hallo" instead of version because I changed the string in my en_US locale and reloaded the extension to show the new localization).

We should update the locale string however to support languages with different word orders. This means using placeholders (https://github.com/EFForg/privacybadger/issues/1329), which I'm not sure our current localization setup supports. This isn't a blocker though.

ghostwords:
Also, you could set up your editor to do linting locally and live as you develop. To start with, you could just install the linter and run it manually on the codebase. From the Badger git checkout folder, run `npm install`, then `make lint`.

src/skin/popup.html
ynasser: Upon restarting my browser, the localization string "version" started being displayed. I guess it was working all along, but wasn't updating locally for me for some reason.

Regarding your suggested changes, I reduced the size of "Privacy Badger" and moved it up a bit, and aligned the version string along the left edge. I've only used CSS and handful of times before and couldn't figure out how to move the version string up. I tried the same technique I used to move "Privacy Badger" up (adjusting the margin), but it seems to be stuck at this height and won't go any higher:

pb_css

ghostwords: You should be able to move title/subtitle strings with `position: relative` and (negative?) `top` values. Though, could you go over this layout with someone who is more familiar with CSS than us? We are currently floating left a bunch of elements and we want to have these two text rows span the logo:
┌─── ┬────┐
│ logo ├────┤
└─── ┴────┘

ynasser: Hm, I didn't see your comment before I came back to work on this, and it turned out the missing link was including `float` for `version`. Does this most recent version look OK:

screenshot from 2017-05-01 21-31-54

ghostwords: That's great! Maybe tweak the amounts of white space a little (add some distance between the logo and the text).

ghostwords:
The [failing job](https://travis-ci.org/EFForg/privacybadger/jobs/227820241) looks like an occasional bug in our Firefox test setup.

Could you check up on a few things before we ship?

- Does this look the same in Firefox?
- Does this make either https://github.com/EFForg/privacybadger/issues/1077 or https://github.com/EFForg/privacybadger/issues/1113 worse?
- What happens if the word for "version" in your language is pretty long? How long is too long?
- Do we need to add the new localization string to all locales?

src/skin/popup.html
ynasser: How does this look:
pb_moved_to_right

There were moved 5px to the right.

ynasser:
> Does this look the same in Firefox?

Yep

> Does this make either #1077 or #1113 worse?
   
I couldn't reproduce those, so I can't confirm that it doesn't make them worse, but my intuition would be no, since we moved the "Privacy Badger" text up to adhere to this layout, and so nothing got pushed down.
┌─── ┬────┐
│ logo ├────┤
└─── ┴────┘


> What happens if the word for "version" in your language is pretty long? How long is too long?

I made the string very, very long to check, and it wraps nicely:

pb_nice_wrap

And long without wrapping:

pb_long_version

> Do we need to add the new localization string to all locales?

It's not required for the app to run, as we have a default locale specified in `manifest.json`. From chrome's i18n documentation: "You don't have to define every string for every supported locale. As long as the default locale's messages.json file has a value for every string, your extension or app will run no matter how sparse a translation is."

We should eventually get this translated into other languages, but I have no idea how we handle that now. Should I make an issue for this?

ghostwords:
Looks good, thank you!

Re translations: I don't think so, although I'm not sure myself what the process is (https://github.com/EFForg/privacybadger/issues/1280).

PR #1362, opened 2017-05-03
stop appending undefined to the return value of getActionDescription (#1092)

### LINKED ISSUE #1092: If DNT policy set, hover on domain shows "undefinedexample.com"
image

Chrome 55.0.2883.87 (64-bit) on Linux Mint 18

ynasser:
Fixes #1092.

`undefined` was sometimes returned when looking up values in the `actionDescriptions` object, and now there's a check for that behaviour.


ynasser:
I've localized the noaction + the dnt action tooltip. However, the way it's currently implemented, there are two tooltips when you hover over the dnt checkmark image (stacked on top of each other):

two_tooltips

The top one says "This domain promises not to track you." and the bottom one repeats the URL that is directly to the right of the image. Is this how it's supposed to be?

ynasser:
I combined the two tooltips and here's what it looks like now:
screenshot from 2017-05-15 11-31-06

I did this by checking if `isWhiteListed` is set, then manually assigning `actionDescription` if so, as this was effectively the old behaviour we had. This is a bit confusing to me though because a site being white listed is different than a site respecting DNT.

src/_locales/en_US/messages.json
```diff
@@ -20,6 +20,12 @@
     "disabled_for_these_domains": {
         "message": "Privacy Badger is disabled for the domains listed below. You can add a domain either via the Privacy Badger icon in the address bar or by typing the domain (not a URL) in the box and clicking the \"Add domain\" button."
     },
+    "dnt_noaction": {
```
ghostwords: Should rename this key since it's not about DNT. `badger_status_noaction` to fit the naming scheme? Oh huh, that one's taken by what should be `badger_status_allow`. Let's fix both.

src/_locales/en_US/messages.json
```diff
@@ -20,6 +20,12 @@
...
+        "message": "No tracking for "
+    },
+    "dnt_tooltip": {
+        "message": "This domain promises to not track you."
```
ghostwords: What's more readable/idiomatic/correct, "to not track" or "not to track"?

src/js/htmlutils.js
```diff
@@ -61,18 +61,20 @@ var htmlUtils = exports.htmlUtils = {
    * @param {String} origin The origin to get description for.
    * @returns {String} Localized action description with origin.
    */
-  getActionDescription: (function () {
```
ghostwords: Looks like we had a bad merge here. Could you restore my changes from 7fddc74cba7d5e5d8721c4bfb2947a726d7092c2?

src/js/htmlutils.js
```diff
@@ -143,11 +145,16 @@ var htmlUtils = exports.htmlUtils = {
...
     var whitelistedText = '';
+    var actionDescription = '';
     if (isWhitelisted) {
       whitelistedText = '' +
         '<div id="dnt-compliant">' +
         '<a target=_blank href="https://www.eff.org/privacybadger#faq--I-am-an-online-advertising-/-tracking-company.--How-do-I-stop-Privacy-Badger-from-blocking-me?">' +
-        '<img src="/icons/dnt-16.png" title="This domain promises not to track you."></a></div>';
+        '<img src="/icons/dnt-16.png"></a></div>';
+
+      actionDescription = i18n.getMessage('dnt_tooltip');
```
ghostwords: We should make this logic part of `getActionDescription`, since it seems like the logic belongs there conceptually.

ghostwords: If getting the description for DNT domains is part of `getActionDescription`, it will also benefit from the caching introduced in https://github.com/EFForg/privacybadger/commit/7fddc74cba7d5e5d8721c4bfb2947a726d7092c2 for example.

src/_locales/en_US/messages.json
ghostwords: Also should we or should we not have a period here? What's most consistent?

ynasser: I think this way is more readable. Also, having periods at the end of full sentences looks more professional, so I'd prefer to keep it.

src/_locales/en_US/messages.json
ynasser: When I update these, do I need to also go update every language other than `en_US`? I'm not sure how the translation stuff works.

ghostwords: Yes, if we're modifying a key, we need to do it everywhere or we'll end up orphaning existing translations.

ynasser:
@ghostwords 

I believe I've addressed all the changes you requested. Here's a summary:
* Added a space between the DNT icon and the domain (5px)
* Updated the options page to include the DNT icon where relevant, including in filtered results
* Updated non-English locales so that `badger_status_noaction` -> `badger_status_allow`, and in English I additionally changed `dnt_noaction`->`badger_status_noaction`.
* Undid the bad merge of `getActionDescriptions`
* Moved some logic that was happening outside of `getActionDescription` that belonged there into it
* Also, I changed my mind about the period at the end of the sentence. Consistency does seem a bit better in this case!

All tests pass locally for me, and I've manually checked the extension on several sites.

ghostwords:
Excellent, thank you!

PR #1425, opened 2017-06-09
WIP: refactoring tracking tabs

cowlicks:
The chrome webrequest api looks like:
```javascript
chrome.webRequest.onBeforeRequest.addListener(function callback)
```
Where the `callback` has to *return* some value to modify the request. Because we need to *synchronously* return something we can't directly use info from *asynchronous* apis like `tabs` or `storage`.  But we often need to associate requests with the url of the tab they are called from, or get info from storage. So we keep track of data from tabs and data from storage in memory so that we can access them synchronously. This pull request addresses problems with how we account for tabs in the `tabData` datastructure.


cowlicks:
This is stale, closing.

PR #1427, opened 2017-06-10
WIP: Refactor how we update the badge in the browser.

cowlicks:
While untangling stuff that refers to `tabData` I refactored some of the code around the badge  into its own module.

Also we were not using a bunch of the icons, and now we are.


cowlicks:
I removed the icons with the badger buddy inside the red hexagon.

cowlicks:
This should be closed in favor of #1480, but I want to move the icon stuff into a separate PR first.

cowlicks:
Succeeded by #1535 and #1480

PR #1461, opened 2017-06-28
Flat UI

koops76:


ghostwords:
The "before" screenshot":

screenshot from 2017-06-28 16 47 29

ghostwords:
The "after" screenshot:

screenshot from 2017-06-28 16 42 20

cowlicks:
I'm a fan! Would it also be possible to remove the shadows under the tooltips? If not, no worries I'm fine merging as-is. Here is a gif of what I'm talking about:
peek 2017-06-28 15-06

koops76:
@cowlicks Done.

koops76:
Should I also remove the slider handles (the =s)?

ghostwords:
Here is what the header tooltips look like now:

screenshot from 2017-06-29 15 43 21

Should we vertically-align them better with the header icons?

ghostwords:
I like these tweaks in that they make the UI more consistent. Some elements (corner and header icons) are already "flat" and some aren't. This makes more elements "flat".

I am concerned that the new sliders are harder to recognize as usable controls, however. The toggles seem to stand out less now.

koops76:
@ghostwords Then I will keep the =s. It would make it easier to notice that they can be interacted with.

andresbase:
What happens if we leave some space between the switch (the red/yellow/green) and the slider (the gray area),like a pixel or so. Would that make it easier to recognize?

Kind of like this:
Image of slider

koops76:
@andresbase I think having a handle is enough to show they can be dragged. Maybe make the handle vertical instead of horizontal like your example?

andresbase:
Considering the size of the switch I think this looks good. Anyone else has thoughts on this?

ghostwords:
I am still concerned that the new sliders are harder to recognize as usable controls since the "flat" toggles seem to stand out less from other elements in the popup.

We've gotten compliments about our current "candy" UI, but no complaints that I know of. Perhaps having header icons "flat" and the toggles shaded is better for usability as the visual differences emphasize what is and isn't clickable.

@andresbase We could try to find the original designer of our toggles and get their feedback.

cowlicks:
I have not complained, but I've personally noted that our UI could use a style change or update to a more modern flat style.

koops76:
@ghostwords The original designer is @ghinda. The switches used are from ghinda/css-toggle-switch.

ghinda:
Hi guys,

Awesome that you're using css-toggle-switch.

Couple of things you can do to make them more visible:
* Like @koops76 said, rotate the handle icon, since they're moving horizontally
* Tweak the colors a bit to make the icon more visible
* Like @andresbase said, 1px padding around the handle
* Slight semi-transparent white top border on the handle. Less "flat" but more visible.
* Remove rounded borders only on the handle. The container still has them.
* +2px extra height.

Here's how it would look:
pb-switch-tweaks

@koops76 I can push the tweaks to your branch, if you can give me access to your fork.

koops76:
@ghinda Done.

ghinda:
@koops76 Thanks. Pushed the changes.

koops76:
@ghinda I'm not sure about:
- Slight semi-transparent white top border on the handle. Less "flat" but more visible.
- Remove rounded borders only on the handle. The container still has them.

Please make a screenshot without them so it can be compared with the current one.

ghinda:
With rounded borders and no top border on the handle.

pb-switch-tweaks-rounded

koops76:
@ghinda Much better! Now please make a screenshot with a white border.

cowlicks:
popup-ui

Currently the UI is like this. @ghostwords do you still have any objections?

koops76:
@cowlicks I updated the CSS, please remake the video.

andresbase:
pbuiflat

koops76:
@andresbase Great, OK to merge.

andresbase:
@koops76 @ghinda thank you for your contributions!

ghostwords:
This is great, thank you @koops76 and @ghinda!

I'm afraid we merged this a bit prematurely though. There are a few issues we should address (that I am aware of so far, still need to review this some more), probably in a followup pull request:
- The spacing on the options page. The taller sliders make the tooltips run into the rows. Could we try making the sliders the same height as the older sliders?
- Is it me, or are the colors of toggles different between the top tracking domains section and the "these domains do not appear to be tracking" sections?
- The header tooltips (that we removed drop shadows from) are misaligned vertically.

ghostwords:
Additionally, the screenshots in the first-time walkthrough (open popup and click on the "?" in the upper right) show the old toggles.

PR #1486, opened 2017-07-12
WIP Quick changes to Flat UI (#1461)

andresbase:
Quick fixes for #1461 
* [x] Same greens for toggles in lists.
* [x] Size of hints when going into options so they don't run into rows
* [ ] Vertical alignment of toolbox (where shadow was removed).

Last one actually is aligned but looks weird so I changed the padding on the top from 1 to 3px. It looks like this:

comppb_padding

### DIFF
--- a/src/skin/popup.css
+++ b/src/skin/popup.css
@@ -76,7 +76,7 @@ a:hover { color: #ec9329 }
 .clicker{
   border-top: 1px solid #aaaaaa;
   padding: 5px;
-  height: 40px;
+  height: 43px;
   overflow: hidden;
 }
 .clicker.greyed { background: #aaa; }
@@ -120,7 +120,7 @@ font-size: 12px;
 background-color: #555;
 height: 12px;
 border-radius: 7px;
-padding: 3px;
+padding: 1px 3px 3px 3px;
 position: relative;
 bottom: 30px;
 left: 0;
--- a/src/skin/toggle-switch.css
+++ b/src/skin/toggle-switch.css
@@ -315,7 +315,7 @@
 }
 
 .noaction .switch-candy a {
-  background-color: #339900;
+  background-color: #4CAF50;
 }
 
 .switch-candy {


ghostwords:
Got a couple of more items to add to the list:
- The undo arrows are misaligned with the toggles (popup and options page)
- The screenshots in the first-time walkthrough (open popup and click on the "?" in the upper right) show the old toggles

andresbase:
Added the alignment for the arrows as it's just one change.
The old message is gone too.


alignarrow

ghostwords:
Could you rebase your branch from master, so that you have the latest patch to ignore Firefox ESR?

A bunch of other tests are failing, for example, `test_cookie_tracker_detection` [here](https://travis-ci.org/EFForg/privacybadger/jobs/253423299). The test may need to be updated, not sure.

ghostwords:
I like the old height (14px) of the slider toggles more.

andresbase:
The toolbox is now aligned, there is an issue with it's alignment on the options page that has been there for a while. Not sure if I'll have time but will try before the release.

toolbox


I tried different sizes these days for the toggles in different screens/resolutions. I think the new size fits better for most screens. 14 looks wrong even if I fix all the other heights, it just ends up looking crammed. 

togglesizes

ghostwords:
The current production design:

current-design

The new toggles in master right now:

new-sliders

The same toggles reduced to current design's height:

new-sliders-current-height

Current design with proposed adjusted-height update next to each other:

side-by-side

ghostwords:
I prefer the new toggles reduced to current design's height.

@koops76 @ghinda Any feedback regarding https://github.com/EFForg/privacybadger/pull/1486#issuecomment-315632063?

ghinda:
I think removing the rounded borders on the handle would look slightly better.

eg.
pb-no-radius

Other than that, looks good.

ghostwords:
I reverted the updated toggles PR merge from master in 4164b3d37310154dfd72c9e0d02ef6942f7f7121 as we still have some fixes to make before releasing the UI updates.

PR #1505, opened 2017-07-16
WIP: Add new heuristic for fingerprinting detection.

cowlicks:
We have list of methods commonly used by browser fingerprinting tools. We wrap each of these in a counter to track their usage. When an origin uses more than some threshold percentage (currently %75) of the methods, we say it is fingerprinting. 

I set the %75 value empirically from testing. The highest percentage I saw on a non-fingerprinting site was 45% (github), the lowest I saw on a fingerprinting site was 80% (this site has since gone down).

There are currently some architectural issue with privacy badger which should be fixed before merging this. Mainly: This kind of tracking should *not* be associated with an **origin**, instead it should be associated with a URL (or some notion of it). It makes sense for cookies to  be associated with an origin, because that is how cookies are scoped. But this tracking is normally loaded from a standalone library that could be anywhere, like a third party CDN service or the first party.

If we block origins associated with fingerprinting, we'll end up blocking CDN's, which breaks other sites, which means we'll manually have to add the sites to the cookieblock list, which will just allow the fingerprinting to continue. I'll expand more on this in a separate issue.

### DIFF
--- a/src/js/contentscripts/fingerprinting.js
+++ b/src/js/contentscripts/fingerprinting.js
@@ -1,3 +1,4 @@
+(function() {
 /*
  * This file is part of Privacy Badger <https://www.eff.org/privacybadger>
  * Copyright (C) 2015 Electronic Frontier Foundation
@@ -294,3 +295,4 @@ document.addEventListener(event_id, function (e) {
 insertFpScript(getFpPageScript(), {
   event_id: event_id
 });
+})();
--- a/src/js/contentscripts/injector.js
+++ b/src/js/contentscripts/injector.js
@@ -1,6 +1,14 @@
-var s = document.createElement('script');
+let event_id = Math.random();
+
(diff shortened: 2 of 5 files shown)


vecna:
image

Hi @cowlicks, I merged this branch in my fork, I'll just report in this PR the issues I encounter

cowlicks:
Hi @vecna ! Awesome, I'm glad someone is testing it out! Note that it currently does not account for domains that were found fingerprinting. The code [here](https://github.com/EFForg/privacybadger/pull/1505/files#diff-f0450b33f099abd6026a2de089706c2cR671) would need to be expanded to register the fingerprinting resource.

cowlicks:
What browser are you using, and what version? What happens if you run this in a the page:
```javascript
Object.defineProperty(window, 'sessionStorage', {get: function() {console.log('stuff')}});
window.sessionStorage
```

PR #1516, opened 2017-07-19
Added support for firefox on android

lemnis:
Tested in Firefox for Android 56.0a1 (2017-07-18), should work from 55+. Did check chrome and firefox desktop, didn't see anything unexpected.

See for the full conversation #1084


cowlicks:
This is awesome @lemnis ! I'll test it out tomorrow.

src/skin/popup.css
```diff
@@ -19,16 +19,20 @@
...
   max-width: 100%;
   font-size: 12px;
   background: #fefefe;
   color: #333;
   font-family: helvetica, arial, sans-serif;
   padding: 8px 15px;
-  width: 400px;
   box-sizing: border-box;
 }
+@media screen and (min--moz-device-pixel-ratio:0) {
```
lemnis: firefox specific css, otherwise the popup would render too wide

src/skin/popup.css
```diff
@@ -19,16 +19,20 @@
...
   margin: 0;
-  min-width: 200px;
+  min-width: 400px;
   max-width: 100%;
   font-size: 12px;
   background: #fefefe;
   color: #333;
   font-family: helvetica, arial, sans-serif;
   padding: 8px 15px;
-  width: 400px;
```
lemnis: creates a whitespace within chrome, use min-width instead

cowlicks:
@lemnis would you mind sharing your workflow for installing this on mobile? I haven't done this before.

lemnis:
1. On your phone, open Firefox and go to `about:config` 
2. Change `xpinstall.signature.required` to true
3. Build the app on your computer, I personally use `web-ext build`
4. Change the file extension from .zip to .xpi
5. Make the .xpi accessible from the phone, use one of the methods below:
    - Use `python -m SimpleHTTPServer` on your computer and open the server with your phone
    - Use `adb push` and open the file location on your phone
6. For debugging, use the Firefox webIDE. All the logs will be shown within the Main Process

Remarks:
- v55+ has a minimum support of `browser_action` that is needed
- I couldn't install the extension on the current beta-build (55.0b10), got a message that the extension appears to be corrupt. Ignore this version...
- Within FF nightly 56.0a1 theirs an issue that after install, the menu item appears 2x times or/and when you press on it, it redirects you to the "report site issue" page. If this happens force shut down the app, after a fresh start the app will behave as expected.

It's a hard life to live on the cutting edge 🙁

lemnis:
All changes in `src/skin/popup/css` should resolve all zoom and unscrollable issues, but could create a vertical scrollbar if needed. Maybe it's best to wait for #1445

cowlicks:
Some extra steps I needed. After transferring the file to my device:
* grant storage permissions to firefox
* in firefox navigate to the url `file:///mnt/sdcard` this should open a file browser, select the xpi file

I'm reviewing now

cowlicks:
screenshot_20170720-172953

cowlicks:
screenshot_20170720-173756

cowlicks:
related to #1276 I ran the pbtest.org tests a few months ago this way and they all failed

This is great! I'll take a closer look at the changes asap. Thank you @lemnis

src/js/popup.js
```diff
@@ -503,6 +503,26 @@ function syncUISelections() {
...
+* @param  {String} url
+* @return {String}     Value of the key
+*/
+function getParameterByName(key, url) {
+  if (!url) { url = window.location.href; }
+  key = key.replace(/[\[\]]/g, "\\$&");
+  var regex = new RegExp("[?&]" + key + "(=([^&#]*)|&|#|$)"), results = regex.exec(url);
+  if (!results) { return null; }
+  if (!results[2]) { return ''; }
+  return decodeURIComponent(results[2].replace(/\+/g, " "));
```
cowlicks: ah I see what this is for

src/js/background.js
```diff
@@ -776,6 +780,25 @@ function startBackgroundListeners() {
...
+      chrome.tabs.query({active: true, lastFocusedWindow: true}, (tabs) => {
+        var url = chrome.runtime.getManifest().browser_action.default_popup + "?id=" + tabs[0].id;
+
+        // check if popup already is open
+        chrome.tabs.query({url}, (popupTabs) => {
+          if(popupTabs.length > 0){
+            chrome.tabs.update(popupTabs[0].id, {active: true});
+            chrome.tabs.reload(popupTabs[0].id);
+          } else {
+            chrome.tabs.create({url, index: tabs[0].index + 1});
```
cowlicks: I think we can get around the passing a url parameter. After you load the popup, you could try doing:
```javascript
chrome.tabs.executeScript(tabId, {code: 'getTab = callback => {callback(' + tabId + ')}'});
```

cowlicks: This would override the `getTab` function in popup with something that works.

cowlicks: Actually, I don't think that would work... Another idea is using `sendMessage`. But if this is working well there is no need to fix it.

src/js/background.js
```diff
@@ -532,16 +530,22 @@ Badger.prototype = {
...
-    } else {
-      chrome.browserAction.setBadgeBackgroundColor({tabId: tabId, color: "#cc0000"});
+    if(chrome.browserAction.setBadgeBackgroundColor){
+      if(numBlocked === 0){
+        chrome.browserAction.setBadgeBackgroundColor({tabId: tabId, color: "#00cc00"});
+      } else {
+        chrome.browserAction.setBadgeBackgroundColor({tabId: tabId, color: "#cc0000"});
+      }
+    }
+    if(chrome.browserAction.setBadgeText){
```
cowlicks: Could we move all of these `if (chrome.browserAction....) {` guards to the top of this function?

Also, could we combine these into a global variable like `isMobile`?

src/js/background.js
```diff
@@ -55,11 +55,9 @@ function Badger() {
...
 
     // Show icon as page action for all tabs that already exist
-    chrome.windows.getAll({populate: true}, function (windows) {
-      for (var i = 0; i < windows.length; i++) {
-        for (var j = 0; j < windows[i].tabs.length; j++) {
-          badger.refreshIconAndContextMenu(windows[i].tabs[j]);
-        }
+    chrome.tabs.query({}, function (tabs) {
+      for (var i = 0; i < tabs.length; i++) {
+        badger.refreshIconAndContextMenu(tabs[i]);
```
cowlicks: 👍

src/js/background.js
lemnis: Moved feature checks.

For what would you use that global variable? All my checks are just feature checks. The best way to determine if it's mobile is `chrome.runtime.getPlatformInfo((platformInfo) => {console.log(platformInfo.os)});`, but i don't see a specific use case.

src/js/background.js
lemnis: I agree there should be a neater solution, what do you think is the best behavior?

- the current, update an existing popup with same parent tab or create a new popup
- always create a new popup and close the popup directly after it was deactivated
- keep at max 1 popup open and update its content

src/js/background.js
cowlicks: The use case is to isolate browser specific code into its own region so developers can reason about it more easily.

For example, I'm imagining someone might be trying to fix some bug on chrome, if they come across this code, it might not be obvious that this intended for use on mobile.

Eventually, it would nice if browser specific code was isolated to independent files, then we could have a shim API. That is out of scope for this PR. But to make it easier to do in the future, it would be nice to have browser specific pieces clearly distinguished.

src/js/background.js
cowlicks: I think that option 2 is the best! It also mimics the behavior of current popups (it isn't long running).  And adding code for deactivating the popup should be simpler than adding the updating code for 3. And I think it will avoid the confusion a user might have when they switch to a popup, and don't know what tab the popup is for.

src/js/background.js
cowlicks: re: way to determine if it is mobile

I think we should use feature existence as a way to determine if its a mobile browser, so we could do something like:

```javascript
let isMobile = !(chrome.browserAction.getPopup && chrome.browserAction.setBadgeText && etc);
```

This also means that we won't have to consider edge cases where some of these API's may exist while others do not.

The downside is we won't immediately get new features if these api's are added. But I think think that is okay. And when they are added all we have to do is remove them from the `isMobile` check.

cowlicks:
Sorry for the delay! I left a few comments, I haven't thought of a better way to pass the id to the popup. I'll let you know if I do, but I think a URL parameter is fine for now. It might also make some selenium tests easier.

src/js/popup.js
```diff
@@ -503,6 +503,26 @@ function syncUISelections() {
...
+*
+* Source: https://stackoverflow.com/a/901144/2662749
+* Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1330159
+*
+* @param  {String} key Key to get from url
+* @param  {String} url
+* @return {String}     Value of the key
+*/
+function getParameterByName(key, url) {
+  if (!url) { url = window.location.href; }
```
cowlicks: I think we can use the [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL) to do this nicely:

```javascript
getTabFromUrl(url) {
  return parseInt(new URL(url).searchParams.get('tabId'));
}
```

The nice thing about working on web extensions is that we only run on newer browsers, so we can use newer API's.

lemnis: I used that API before but didn't bother to read. Changed it know, should have read it before though 💤

src/js/background.js
lemnis: added the boolean to the badger constructor, it's only the browserAction API where it is checked against. 

Added in another pull request other browser specific code 😞 , bloating the source code 😇

lemnis:
Changed the popup behavior.

We could switch to `sendMessage` or `executeScript` because now it's only needed to transfer the parent id, but this would probably be slower.

Also, I was convinced that there must be a chromium-based android browser that supports extensions. After some searches I found that the Yandex browser supports it, they use the opera webstore. So Privacy Badger is really easy to install and it already works! Just wanted to let you know, if you didn't know already.

cowlicks:
I'm testing this out again today.

src/js/background.js
```diff
@@ -776,6 +781,29 @@ function startBackgroundListeners() {
...
+        var url = chrome.runtime.getManifest().browser_action.default_popup + "?tabId=" + tabs[0].id;
+        chrome.tabs.create({url, index: tabs[0].index + 1}, (tab) => {
+          openPopupId = tab.id;
+        });
+      });
+    });
+
+    // remove the 'popup' when another tab is activated
+    chrome.tabs.onActivated.addListener((activeInfo) => {
+      if(openPopupId && openPopupId != activeInfo.tabId) {
```
cowlicks: I think this will fail if `openPopupId = 0`. We can instead check for a sentinel value like:
```javascript
var openPopup = false;
...
if ((openPopupId != false) && ...) {
  ...
  openPopupId = false;
}
```

Also, I think there might be an issue here if the user changes the url of the popup tab to something else, then visits another tab. Would it close the tab that was formerly a popup? I'll try this out when I'm testing today.

cowlicks:
Thanks for the update @lemnis, sorry for the delay. When I last checked this out I was having problems with the popup. I'll try it out again today.

lemnis:
We are both slow, untill nobody is unpeasantly waiting don't feel pressured to exuse yourself or keep up to a certain paise. As a dutch girl, I'm not a 'mierenneuker' 😉 

The code is updated so that a former popup will not close if the user changed the url. I doubted if the code for `onUpdated` should be in his own eventlistener or added to to the current. I have eventually choosen the second.

As of [bugzilla#1386786](https://bugzilla.mozilla.org/show_bug.cgi?id=1386786) is fixed, we can decide to add the number of blocked resources to the title between parenthesis. I removed the line that did set the title within this pull request [background.js:765](https://github.com/EFForg/privacybadger/pull/1516/files#diff-bc3b0204e92c9ade638a46f86d54e8c8L765), because it was a fixed string of text. 

There are 3 options:
* Just don't bother to add the code
* Only set a 'custom' title for FF android
* Set a 'custom' title for every browser, so when you hover over the icon, you can read the amount of blocked resources

The benefit would maybe dissappear for FF mobile in the future, if mozilla decides to use icons. Also this added information could benefit blind users, but I don't know if the current plugin is usable for them.

src/js/background.js
```diff
@@ -762,7 +768,6 @@ Badger.prototype = {
     }
 
     chrome.browserAction.setIcon({tabId: tab.id, path: iconFilename});
-    chrome.browserAction.setTitle({tabId: tab.id, title: "Privacy Badger"});
```
cowlicks: Nice, now this is localized properly by the `"default_title"` key in the manifest

src/js/popup.js
```diff
@@ -512,6 +512,14 @@ function syncUISelections() {
 * seems to be that it is synchronous.
 */
 function getTab(callback) {
+  // Temporary fix for Firefox Android while it doesn't support `browser_action.default_popup`
+  if(badger.isFirefoxMobile){
+    chrome.tabs.query({active: true, lastFocusedWindow: true}, function(focusedTab) {
```
cowlicks: I think we can use `chrome.tabs.getCurrent` here instead

src/js/background.js
```diff
@@ -776,12 +781,45 @@ function startBackgroundListeners() {
     }
   }, {urls: ["http://*/*", "https://*/*"]}, []);
 
+  if(badger.isFirefoxMobile) {
```
cowlicks: Could you refactor these to be like:
```javascript
if (!isMobile) {
  startRegularListeners(); // with an exposed named function
} else {
 startMobileListeners();
}
```

This way the mobile logic is separate from the regular logic. And so we are able to test them independently .

src/js/popup.js
lemnis: `getCurrent` unfortunately doesn't work

lemnis:
(Finally) moved all the code to a separate file. Maybe not the neatest code, because I am putting functions for both the popup.js and the background.js inside the same file. Besides that, feel free to change the function names, the ones that I have chosen are maybe not the best for knowing where you should use the functions.

cowlicks:
To fix the tests you just need to include the firefoxandroid.js file in `src/tests/index.html`. Otherwise LGTM

cowlicks:
@lemnis great work! Thank you.

strugee:
Does this mean that Android support is "done"/stable? Or just much closer?

Context: nylira/prism-break#1648

cowlicks:
@strugee "just much closer" I think. It hasn't been tested much. And I think we should consider revamping some UI before we officially support it.

strugee:
:thumbsup: thanks for clarifying!

ghostwords:
The domain list key/legend icons seem misaligned in the popup and the options page:
screenshot from 2017-08-21 14 25 54.

This is coincidentally fixed in my WIP flexible popup layout PR (#1445) since I needed to change how the key/legend icons get rendered.

ghostwords:
It seems that we also introduced a body scrollbar in case of long-enough domain lists in Firefox:

screenshot from 2017-08-22 17 03 07

lemnis:
@ghostwords https://github.com/EFForg/privacybadger/pull/1516#issuecomment-316726947 😉 

But, maybe adding temporarily `body { overflow: hidden; }` wouldn't be a terrible idea, it will remove the scrollbars and nobody will notice it besides the single user that already uses Privacybadger on their phone (a.k.a. me)

I overlooked the misalignment of the key/legend icons...

cowlicks:
@ghostwords would you mind creating a separate issue so this can be more easily tracked?

PR #1535, opened 2017-07-27
Remove extra unused icons.

cowlicks:
Taken from #1427


cowlicks:
They can sometimes get used when the scale is appropriate.. I remember seeing it once on firefox. However if you look at the image, it is the old icon with the red hexagon that we don't use anymore. The other icons will get scaled when appropriate.

PR #1739, opened 2017-10-18
Stop errors from leaking to the console

ddsol:
The local storage clobbering can cause security errors when third-party cookies are being blocked. Previously, the error would leak out onto the console producing noise that makes debugging sites harder. This commit catches the error and ignores it.

### DIFF
--- a/src/js/contentscripts/clobberlocalstorage.js
+++ b/src/js/contentscripts/clobberlocalstorage.js
@@ -48,7 +48,8 @@ chrome.runtime.sendMessage({checkLocation:document.location.href}, function(bloc
           window.localStorage.setItem = function(/*newValue*/) {
             //doNothing
           };
-        } catch(ex) {
+        } catch (ex) {
+          // ignore error
         }
       } +')()';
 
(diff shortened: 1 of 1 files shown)


ghostwords:
Hello! Could you provide an example situation when this happens? What should I do to see one of these errors? What do they look like?

ghostwords:
Could you also address the linter complaints? You can resolve the "no empty blocks" error by adding an explanatory comment inside the catch.

ddsol:
When this is set in Chrome:
block cookies
there's an error when the `localStorage` is being clobbered:

```
Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
```

ghostwords:
Ah! This came up in the past in a different injected script: #433.

ghostwords:
Documented here: https://www.chromium.org/for-testers/bug-reporting-guidelines/uncaught-securityerror-failed-to-read-the-localstorage-property-from-window-access-is-denied-for-this-document.

>This exception is thrown when the "Block third-party cookies and site data" checkbox is set in Content Settings. ... If this setting is checked, third-party scripts cookies are disallowed and access to localStorage may result in thrown SecurityError exceptions.

ghostwords:
Thank you! I updated the comment in 670f857a8a8b91bf235085459b4124931a1c3c7a.

PR #1758, opened 2017-11-01
issue #1743: Enhance the tutorial message in popup

ymhuang0808:
The pull request fix issue #1743 . Please take a look.

It contains the following changes for comic in popup:
- The instruction message doesn't show when the current page is firstRun.html.
- If the first run link in popup is clicked, it switched into existing firstRun.html tab or create a new tab.

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -59,15 +59,30 @@ function showNagMaybe() {
   }
 
   if (!seenComic) {
-    nag.show();
-    outer.show();
-    // Attach event listeners
-    $('#fittslaw').click(_hideNag);
-    $("#firstRun").click(function() {
-      chrome.tabs.create({
-        url: chrome.extension.getURL("/skin/firstRun.html#slideshow")
-      });
-      _hideNag();
+    chrome.tabs.query({active: true, currentWindow: true}, function (focusedTab) {
+      var firstRunUrl = chrome.extension.getURL("/skin/firstRun.html");
+
+      // Show the popup instruction if the active tab is not firstRun.html page
+      if (!focusedTab[0].url.startsWith(firstRunUrl)) {
+        nag.show();
+        outer.show();
+        // Attach event listeners
+        $('#fittslaw').click(_hideNag);
+        $("#firstRun").click(function() {
+          // If there is a firstRun.html tab, switch to the tab.
+          // Otherwise, create a new tab
+          chrome.tabs.query({url: firstRunUrl}, function (tabs) {
+            if (tabs.length == 0) {
+              chrome.tabs.create({
+                url: chrome.extension.getURL("/skin/firstRun.html#slideshow")
+              });
+            } else {
+              chrome.tabs.update(tabs[0].id, {active: true});
+            }
+            _hideNag();
+          });
+        });
+      }
     });
   }
 }
--- a/tests/selenium/popup_test.py
+++ b/tests/selenium/popup_test.py
@@ -41,8 +41,11 @@ class PopupTest(pbtest.PBSeleniumTest):
 
         self.driver.switch_to.window(new_handle.pop())
 
-    def open_popup(self, close_overlay=True):
+    def open_popup(self, close_overlay=True, new_url=True):
         """Open popup and optionally close overlay."""
+        if new_url:
+            self.load_url("http://eff-tracker-site1-test.s3-website-us-west-2.amazonaws.com")
+
         self.load_url(self.popup_url, wait_on_site=1)
 
         # hack to get tabData populated for the popup's tab
@@ -63,7 +66,7 @@ class PopupTest(pbtest.PBSeleniumTest):
             message="Timed out waiting for getTab() to complete."
         )
 
-        if close_overlay:
+        if close_overlay and new_url:
             # Click 'X' element to close overlay.
             try:
                 close_element = self.driver.find_element_by_id("fittslaw")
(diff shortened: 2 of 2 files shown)


ymhuang0808:
I should fix the tests.

ymhuang0808:
I think the failed test is related with #1634 . In Chrome, the focused tab URL is popup tab instead of the first run page. Should I remove the tests before #1634 is fixed?

BTW, I found the `trackers_link = self.driver.find_element_by_link_text("trackers")`  in `PopupTest::test_trackers_link()` always breaks in my environment. It results from the default language in Chrome. Since, my native language and environment is Traditional Chinese (zh_TW), therefore, Chrome browser is opened with Traditional Chinese in default. On Linux, I should set the `LANGUAGE` env variable when running tests. Reference from https://developer.chrome.com/extensions/i18n#locales-testing 

The command looks like:
```
LANGUAGE=en BROWSER=chrome pytest -v
```

Before:


After:

src/js/popup.js
```diff
@@ -59,15 +59,30 @@ function showNagMaybe() {
...
+      // Show the popup instruction if the active tab is not firstRun.html page
+      if (!focusedTab[0].url.startsWith(firstRunUrl)) {
+        nag.show();
+        outer.show();
+        // Attach event listeners
+        $('#fittslaw').click(_hideNag);
+        $("#firstRun").click(function() {
+          // If there is a firstRun.html tab, switch to the tab.
+          // Otherwise, create a new tab
+          chrome.tabs.query({url: firstRunUrl}, function (tabs) {
```
ghostwords: If I have the tutorial page open in a different window, clicking on the tutorial link doesn't seem to do anything.

ghostwords:
>  I found the trackers_link = self.driver.find_element_by_link_text("trackers") in PopupTest::test_trackers_link() always breaks in my environment.

Ah! Makes sense, the word "trackers" wouldn't be found in the popup on non-English locales. Is this the only locale-related test failure? We should fix this one by making the link acquisition language-agnostic. And, if we went to the trouble of opening the eff.org page, we might as well verify that the page actually contains the anchor element we wanted to scroll to.

tests/selenium/popup_test.py
```diff
@@ -41,8 +41,11 @@ def switch_to_new_window(self, handles_before, max_retries=5):
 
         self.driver.switch_to.window(new_handle.pop())
 
-    def open_popup(self, close_overlay=True):
+    def open_popup(self, close_overlay=True, new_url=True):
         """Open popup and optionally close overlay."""
+        if new_url:
+            self.load_url("http://eff-tracker-site1-test.s3-website-us-west-2.amazonaws.com")
```
ghostwords: Since `new_url` defaults to `True`, this opens http://eff-tracker-site1-test.s3-website-us-west-2.amazonaws.com for most popup tests. Seems like we should instead issue this instruction inside the one test that needs it. I'm also not clear on why we do it; we should explain with a comment.

tests/selenium/popup_test.py
```diff
@@ -244,6 +247,42 @@ def test_donate_button(self):
...
 
+    def test_not_show_instruction(self):
+        """Ensure the instruction message doesn't display on firstRun.html ."""
+        self.open_popup(new_url=False)
+
+        try:
+            instruction = self.driver.find_element_by_id("instruction")
+        except NoSuchElementException:
+            self.fail("Unable to find instruction on popup")
+        self.assertEqual(instruction.value_of_css_property("display"), "none",
```
ghostwords: Yep, this doesn't work because we don't have a proper way to open the popup on/for a particular page in tests (https://github.com/EFForg/privacybadger/issues/1634#issuecomment-338265211). Instead the popup is opened in the context of the popup itself I think. Why did this test pass some of the time in https://travis-ci.org/EFForg/privacybadger/builds/295583717?

ymhuang0808:
Yes, if I doesn't set the language into _en_, the chrome is launched in my system's language. The simplest way is setting the language into _en_, when running testings. 

I don't understand that:

> We should fix this one by making the link acquisition language-agnostic.

Is this like checking a specific selector?

ghostwords:
>Is this like checking a specific selector?

Yes, exactly, I meant to say let's query by something that won't be affected by the locale.

src/js/popup.js
ghostwords: Cool!

We might be going beyond what needs to happen by implementing tab/window switching logic, instead of just unconditionally opening a new tab as we do now. But it is nicer.

tests/selenium/popup_test.py
ghostwords: Don't we call `self.load_url` again directly below? Then don't we always end up on `self.popup_url`, regardless of the value of `new_url`?

ghostwords:
@ymhuang0808 Would you like me to finish this pull request? No rush, just checking in. Let me know if you'd like to make the fixes and look this over again yourself.

ghostwords:
I edited the description to make use of GitHub's [issue-auto-closing syntax](https://help.github.com/articles/closing-issues-using-keywords/).

ymhuang0808:
@ghostwords Yes! Since I'm a little busy this week, could you give me some time to finish the issue?

ghostwords:
Yep, of course! Like I said, no rush. Just checking.

src/js/popup.js
ymhuang0808: The latest commit contains tab and window switching. Please take a look.

ghostwords:
Thank you! Merged with some tweaks.

I fixed a bug with switching tabs within the same browser window, tweaked the code for readability, and removed the WIP tests in a few [followup commits](https://github.com/EFForg/privacybadger/compare/914f783af27e92b760a831de180ac5604c0e9618~3...914f783af27e92b760a831de180ac5604c0e9618).

[Here is the combined changeset](https://github.com/EFForg/privacybadger/compare/914f783af27e92b760a831de180ac5604c0e9618~4...914f783af27e92b760a831de180ac5604c0e9618?w=1) for this PR and my followups, with whitespace ignored.

I fixed the tests when run with a non-English locale in cfb213208f700bacc4f154f70dab6c69b46b165d, and added an extra check for the `eff.org` FAQ page in 68d4cb81a3089db6183ec7d1ada6b8919db5b162.

ghostwords:
Fixed popup tests on Firefox in 636c81b575f66cff16f2f75dfc076e3d2fbd8551.

Added waiting to FAQ check in 2886431fb44789eebc0c6174aae2a652460fd988.

ymhuang0808:
@ghostwords Thanks for your improvement. I'm sorry for breaking something.

ghostwords:
No problem! Selenium tests are already painful to work with, and our popup tests are especially bad since we have to fake opening the popup and the current faking is done badly (to be improved later, perhaps with #1634).

PR #2020, opened 2018-05-14
Error text stays when user clicks away

fr0gs:
Fixes https://github.com/EFForg/privacybadger/issues/1965, sorry I messed up with previous branch.


fr0gs:
I will close  this and create a new PR using that strategy then.

PR #2039, opened 2018-05-27
Unwrap links at Twitter PWA

rugk:
Fixes https://github.com/EFForg/privacybadger/issues/1886

It expands links in the timeline and user bio/profile.

Does not include support for Twitter cards, because I just could not find the URL anywhere in the page. (At least not in the HTML part.)

It may only expand twitter cards, if the link has been used on some other part of the page, but that is not very likely.

### DIFF
--- a/src/js/firstparties/twitter.js
+++ b/src/js/firstparties/twitter.js
@@ -2,7 +2,7 @@ const tcos_all = getTCoSelectorWithDestination();
 const timeline_refresh_interval = 2000;
 
 // twitter.com
-const full_url_attribute = 'data-expanded-url';
+const full_url_attribute = 'title';
 const tcos_with_destination = getTCoSelectorWithDestination(full_url_attribute);
 const full_url_attribute_profile = "title";
 const profile_links_tcos = ".ProfileHeaderCard " + getTCoSelectorWithDestination(full_url_attribute_profile);
@@ -91,6 +91,12 @@ function unwrapTwitterPwaURLsInTimeline() {
       unwrapTco(link, fixes[link.href]);
       return;
     }
+    
+    // use title, if possible
+    if (fixes.hasOwnProperty(link.href)) {
+      unwrapTco(link, fixes[link.href]);
+      return;
+    }
 
     // find span element with link text in it
     const allSpans = link.getElementsByTagName("span");
@@ -144,14 +150,9 @@ function unwrapSpecialTwitterPwaURLs() {
   }
 }
 
-if (window.location.hostname == "mobile.twitter.com") {
-  unwrapSpecialTwitterPwaURLs();
-  unwrapTwitterPwaURLsInTimeline();
+unwrapTwitterURLsInTimeline();
+unwrapSpecialTwitterPwaURLs();
+unwrapTwitterPwaURLsInTimeline();
 
-  setInterval(unwrapTwitterPwaURLsInTimeline, timeline_refresh_interval);
-} else {
-  unwrapTwitterURLsInTimeline();
-  unwrapSpecialTwitterURLs();
-
-  setInterval(unwrapTwitterURLsInTimeline, timeline_refresh_interval);
-}
+setInterval(unwrapTwitterURLsInTimeline, timeline_refresh_interval);
+setInterval(unwrapTwitterPwaURLsInTimeline, timeline_refresh_interval);
--- a/src/js/firstparties/twitter.js
+++ b/src/js/firstparties/twitter.js
@@ -129,14 +129,9 @@ function unwrapSpecialTwitterURLs() {
 }
 
 function unwrapSpecialTwitterPwaURLs() {
-  // profile URL can be found in JSON of page
-  if (!window.wrappedJSObject) {
-    return;
-  }
-
   try {
     // iterate users object (usually should only be one, with some random ID?)
-    for (const entityKey of Object.keys(window.wrappedJSObject.__INITIAL_STATE__.entities.users.entities)) {
+    for (const entityKey of Object.keys(window.__INITIAL_STATE__.entities.users.entities)) {
       const entityValue = window.wrappedJSObject.__INITIAL_STATE__.entities.users.entities[entityKey];
       // iterate url array (usually only one)
       for (const url of entityValue.entities.url.urls) {
@@ -151,8 +146,9 @@ function unwrapSpecialTwitterPwaURLs() {
 }
 
 unwrapTwitterURLsInTimeline();
-unwrapSpecialTwitterPwaURLs();
 unwrapTwitterPwaURLsInTimeline();
 
+// only works at initial load if you directly load a profile
+unwrapSpecialTwitterPwaURLs();
+
 setInterval(unwrapTwitterURLsInTimeline, timeline_refresh_interval);
-setInterval(unwrapTwitterPwaURLsInTimeline, timeline_refresh_interval);
--- a/src/js/firstparties/twitter.js
+++ b/src/js/firstparties/twitter.js
@@ -83,41 +83,6 @@ function unwrapTwitterURLsInTimeline() {
   });
 }
 
-function unwrapTwitterPwaURLsInTimeline() {
-  // unwrap profile links
-  document.querySelectorAll(tcos_all).forEach((link) => {
-    // use cached value if possible
-    if (fixes.hasOwnProperty(link.href)) {
-      unwrapTco(link, fixes[link.href]);
-      return;
-    }
-    
-    // use title, if possible
-    if (fixes.hasOwnProperty(link.href)) {
-      unwrapTco(link, fixes[link.href]);
-      return;
-    }
-
-    // find span element with link text in it
-    const allSpans = link.getElementsByTagName("span");
-    if (!allSpans.length) {
-      return;
-    }
-
-    const linkElement = allSpans[0];
-    const elements = link_reg_ex.exec(linkElement.textContent);
-    if (elements === null) {
-      return;
-    }
-
-    const url = elements[1];
-    if (url) {
-      fixes[link.href] = url;
-      unwrapTco(link, url);
-    }
-  });
-}
-
 function unwrapSpecialTwitterURLs() {
   // unwrap profile links
   document.querySelectorAll(profile_links_tcos).forEach((link) => {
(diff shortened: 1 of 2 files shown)


ghostwords:
Sorry, apologies for the delay, but I haven't forgotten, on my list.

src/js/firstparties/twitter.js
```diff
@@ -1,11 +1,24 @@
+const tcos_all = getTCoSelectorWithDestination();
+const timeline_refresh_intervall = 2000;
```
ghostwords: Typo: "intervall"

src/js/firstparties/twitter.js
```diff
@@ -63,9 +76,33 @@ function unwrapTwitterURLsInTimeline() {
...
+  // unwrap profile links
+  document.querySelectorAll(tcos_all).forEach((link) => {
+    // use cached value if possible
     if (fixes.hasOwnProperty(link.href)) {
       unwrapTco(link, fixes[link.href]);
+      return;
+    }
+
+    // find span element with link text in it
+    const linkElement = link.getElementsByTagName("span")[0];
```
ghostwords: What if there are no `span` elements? Should test to avoid trying to read properties of `undefined` (JS exception).

src/js/firstparties/twitter.js
```diff
@@ -80,6 +117,28 @@ function unwrapSpecialTwitterURLs() {
...
 }
 
-unwrapSpecialTwitterURLs();
-unwrapTwitterURLsInTimeline();
-setInterval(unwrapTwitterURLsInTimeline, 2000);
+function unwrapSpecialTwitterPwaURLs() {
+  // profile URL can be found in JSON of page
+
+  // iterate users object (usually should only be one, with some random ID?)
+  for (const entityKey of Object.keys(window.wrappedJSObject.__INITIAL_STATE__.entities.users.entities)) {
```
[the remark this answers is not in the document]
rugk: Hmm, good catch! So how should I solve that? Just surround it with a catch block and ignore issues?

src/js/firstparties/twitter.js
```diff
@@ -63,13 +76,42 @@ function unwrapTwitterURLsInTimeline() {
...
+      return;
+    }
+
+    // find span element with link text in it
+    const allSpans = link.getElementsByTagName("span");
+    if (!allSpans) {
+      return;
+    }
+
+    const linkElement = link.getElementsByTagName("span")[0];
```
ghostwords: Could we reuse `allSpans` to avoid rerunning the lookup?

src/js/firstparties/twitter.js
```diff
@@ -63,13 +76,42 @@ function unwrapTwitterURLsInTimeline() {
...
+  document.querySelectorAll(tcos_all).forEach((link) => {
+    // use cached value if possible
+    if (fixes.hasOwnProperty(link.href)) {
+      unwrapTco(link, fixes[link.href]);
+      return;
+    }
+
+    // find span element with link text in it
+    const allSpans = link.getElementsByTagName("span");
+    if (!allSpans) {
```
ghostwords: I think we need to test `allSpans.length` here instead. An empty `HTMLCollection` is truthy.

src/js/firstparties/twitter.js
```diff
@@ -80,6 +122,33 @@ function unwrapSpecialTwitterURLs() {
...
 
-unwrapSpecialTwitterURLs();
-unwrapTwitterURLsInTimeline();
-setInterval(unwrapTwitterURLsInTimeline, 2000);
+function unwrapSpecialTwitterPwaURLs() {
+  // profile URL can be found in JSON of page
+
+  try {
+    // iterate users object (usually should only be one, with some random ID?)
+    for (const entityKey of Object.keys(window.wrappedJSObject.__INITIAL_STATE__.entities.users.entities)) {
```
ghostwords: `wrappedJSObject` is Firefox-only. Getting errors in Chrome:
```
TypeError: Cannot read property '__INITIAL_STATE__' of undefined
    at unwrapSpecialTwitterPwaURLs (twitter.js:130)
    at twitter.js:145
```

ghostwords: If there isn't a straightforward cross-browser way to grab the profile link, we might want to just skip profile link replacement on the Twitter PWA.

src/js/firstparties/twitter.js
```diff
@@ -63,13 +76,42 @@ function unwrapTwitterURLsInTimeline() {
...
+    }
+
+    // find span element with link text in it
+    const allSpans = link.getElementsByTagName("span");
+    if (!allSpans) {
+      return;
+    }
+
+    const linkElement = link.getElementsByTagName("span")[0];
+    const elements = link_reg_ex.exec(linkElement.textContent);
```
ghostwords: Getting errors (in Chrome) here, might be related to the empty `HTMLCollection` thing above:
```
twitter.js:102 Uncaught TypeError: Cannot read property 'textContent' of undefined
    at document.querySelectorAll.forEach (twitter.js:102)
    at NodeList.forEach (<anonymous>)
    at unwrapTwitterPwaURLsInTimeline (twitter.js:88)
```

src/js/firstparties/twitter.js
rugk: Oh, of course, missed it.

src/js/firstparties/twitter.js
rugk: Okay, so what is the chrome equivalent?

ghostwords: I don't think there is a direct replacement, and I'm not sure it's worth complicating this script just for the profile link.

Can we get the full URL from the title attribute, same as on regular Twitter? Is the problem that we don't have a stable-seeming selector to target the profile area? But querying for all http/https t.co href attributes with http/https title attributes should pick that link up too, no?

rugk:
Uh, sorry, forgot this one here. Should I still continue on this or is this obsolete?

ghostwords:
@rugk Sorry I missed your message back in April! If you're still interested, let's continue for sure. Let me know. I could also pick up and finish the PR where you left off.

rugk:
Puuh… from what I see Twitter's new PWA is really ugly, because they actively prevent unwrapping these shortlinks.
Anyway, I'll have a look. Please ping me again, if I miss it.

rugk:
(Err, can you assign this PR to me?)

src/js/firstparties/twitter.js
rugk: now just check for `wrappedJSObject` (feature detection), and ignore it if it is not needed

rugk:
Apparently Twitter really fights us here. They even hide the ` title`  attribute for Twitter cards:

image

That said, i could get ti to work again. And I could get rid of `wrappedJSObject`. It seems, in newer Firefox'es this is also not supported… :thinking: 
Anyway, this also only works when the variable is still there.

It's all a little buggy, but it works. At least usual links can be easily expanded.

src/js/firstparties/twitter.js
```diff
@@ -80,6 +93,26 @@ function unwrapSpecialTwitterURLs() {
   });
 }
 
-unwrapSpecialTwitterURLs();
+function unwrapSpecialTwitterPwaURLs() {
+  try {
+    // iterate users object (usually should only be one, with some random ID?)
+    for (const entityKey of Object.keys(window.__INITIAL_STATE__.entities.users.entities)) {
```
ghostwords: Getting an error here: "Cannot read property 'entities' of undefined".

src/js/firstparties/twitter.js
```diff
@@ -80,6 +93,26 @@ function unwrapSpecialTwitterURLs() {
...
+      const entityValue = window.wrappedJSObject.__INITIAL_STATE__.entities.users.entities[entityKey];
```
ghostwords: I don't think we should make use of `wrappedJSObject` at all as it's a non-standard property that's not supported in any modern browser at this point.

ghostwords: If we want to escape the separate execution context that content scripts operate in, we should instead have the content script inject a script into the page ([for example](https://github.com/EFForg/privacybadger/blob/8831ac842100487e15482baa71ade464b3dd7535/src/js/contentscripts/dnt.js#L56)).

rugk: Sooo, looked again at why this is done, and the reason is the XRayVision in Firefox, see https://developer.mozilla.org/en-US/docs/Mozilla/Tech/Xray_vision

And no, that script is already injected. I could not find any documentation for `injectScript` . I guess it's just an old alias for [`executeScript`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/executeScript) or so. It also seems to be some bad eval or so and is not listed in this functions here, you can usually use to clone objects or functions into the site etc: https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Language_Bindings/Components.utils

src/js/firstparties/twitter.js
rugk: Ah yeah, that `wrappedJSONObject` had a reason. And anyway, this is why this is in a try box, if it fails, we don't care. That feature is not important.

rugk:
> Does this variable exist for you on mobile.twitter.com when you are not signed in? When does it exist for you? Does it exist in both Chrome and Firefox?

I guess only Firefox (or could you reproduce it in Chrome), 

> Which URLs does this variable contain? The profile URL? Card URLs? What about URLs of links that load dynamically as you keep scrolling the page?

Only the profile URL.
Only at initial load, not dynamically…

Tested it with loading https://twitter.com/rugkme directly and it worked also when not signed in. (though when not signed in, you get a different site on the desktop, which looks/seems to be the old site/design of Twitter.

So yeah, an edge case and a creepy workaround… :thinking:

src/js/firstparties/twitter.js
ghostwords: [`injectScript()` is defined in another content script](https://github.com/EFForg/privacybadger/blob/18feb68d8a5d50a6209dbd24851283f922db71c2/src/js/contentscripts/utils.js#L24).

ghostwords:
As it appears that we can't unwrap `t.co` links in the new Twitter UI (outside of the profile URL, and maybe only in Firefox), we shouldn't proceed further at this point.

Thank you for looking into this!

rugk:
:cry:

PR #2141, opened 2018-08-21
Add Springer Nature domains to MDFP list

bcyphers:
snitch map from badger-sett data:
```
"springernature.com": [
  "nature.com",
  "biomedcentral.com",
  "springer.com"
],
```


bcyphers:
Added the new domains.
springerhealthcare.com and metzlerverlag.de are from https://group.springernature.com/gp/group/aboutus (under "Our trusted brands"), but I'll remove them if you want.

PR #2142, opened 2018-08-22
Tweak seed data to fix bugs

bcyphers:
Update to the latest version of the seed data from badger-sett.


bcyphers:
On second thought, this is worth merging in to fix inconsistencies.

PR #2150, opened 2018-08-27
Sync disabled sites using chrome.storage.sync

Jsnhlbr5:
This is not remotely done, just opening this to facilitate discussion, per @ghostwords suggestion.

This PR is for issue #2110

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,11 +50,11 @@ tests/.pytest_cache
 
 .idea
 
+#Eclipse project stuff
+.settings/
+.project
+
 web-ext-artifacts
 pkg
 release-utils/config.sh
 release-utils/xpi
-
-#Eclipse project stuff
-.settings/
-.project
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,7 @@ web-ext-artifacts
 pkg
 release-utils/config.sh
 release-utils/xpi
+
+#Eclipse project stuff
+.settings/
+.project
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -510,7 +510,7 @@
         }
     },
     "sync_intro": {
-        "message": "Cloud sync:<ul><li>Requires Firefox/Chrome Sync</li><li>Upload overwrites any existing data in the cloud</li><li>Download combines the lists of sites where your Badger is disabled</li></ul>",
+        "message": "Cloud sync:<ul><li>Requires Firefox/Chrome Sync</li><li>Upload overwrites any existing Privacy Badger data in the cloud</li><li>Download combines the lists of sites where your Badger is disabled</li></ul>",
         "description": "A brief explanation of how syncing works. Shown above the upload/download cloud data buttons under the Manage Data options page tab."
     },
     "upload": {
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -518,11 +518,11 @@
         "description": "Button label for uploading data to the cloud on Manage Data options page tab."
     },
     "upload_cloud": {
-        "message": "Upload disabled sites to cloud",
+        "message": "Export disabled sites to cloud",
         "description": "Button title for uploading data to the cloud on Manage Data options page tab."
     },
     "download_cloud": {
-        "message": "Download disabled sites from cloud",
+        "message": "Import disabled sites from cloud",
         "description": "Button title for downloading data from the cloud on Manage Data options page tab."
     },
     "download_cloud_success": {
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -514,35 +514,35 @@
         }
     },
     "sync_intro": {
-        "message": "Cloud syncing:<ul><li>Requires Firefox/Chrome Sync</li><li>Download combines the lists of sites where your Badger is disabled</li><li>Upload overwrites the list saved in the cloud</li></ul>",
-        "description": "A brief explanation of how syncing works. Shown above the upload/download disabled sites buttons under the Manage Data options page tab."
+        "message": "Cloud sync:<ul><li>Requires Firefox/Chrome Sync</li><li>Upload overwrites any existing data in the cloud</li><li>Download combines the lists of sites where your Badger is disabled</li></ul>",
+        "description": "A brief explanation of how syncing works. Shown above the upload/download cloud data buttons under the Manage Data options page tab."
     },
     "upload": {
         "message": "Upload",
-        "description": "Button label for Sync Upload on Manage Data options page tab."
+        "description": "Button label for uploading data to the cloud on Manage Data options page tab."
     },
-    "upload_whitelist": {
+    "upload_cloud": {
         "message": "Upload disabled sites to cloud",
-        "description": "Button title for Sync Upload on Manage Data options page tab."
+        "description": "Button title for uploading data to the cloud on Manage Data options page tab."
     },
-    "download_whitelist": {
+    "download_cloud": {
         "message": "Download disabled sites from cloud",
-        "description": "Button title for Sync Download on Manage Data options page tab."
+        "description": "Button title for downloading data from the cloud on Manage Data options page tab."
     },
-    "download_whitelist_success": {
+    "download_cloud_success": {
         "message": "Disabled sites list downloaded from cloud and merged into local list.",
-        "description": "Message displayed to the user after successfully downloading the synced whitelist."
+        "description": "Message displayed to the user after successfully merging cloud data."
     },
-    "download_whitelist_failure": {
+    "download_cloud_failure": {
         "message": "Disabled sites list could not be downloaded from cloud.  (You will get this message if you have not uploaded any list to the cloud.)",
-        "description": "Message displayed to the user after failing to download the synced whitelist."
+        "description": "Message displayed to the user after failing to download cloud data."
     },
-    "upload_whitelist_success": {
+    "upload_cloud_success": {
         "message": "Disabled sites list uploaded to cloud.",
-        "description": "Message displayed to the user after successfully uploading the local whitelist."
+        "description": "Message displayed to the user after successfully uploading local data."
     },
-    "upload_whitelist_failure": {
+    "upload_cloud_failure": {
         "message": "Disabled sites list could not be uploaded to cloud.",
-        "description": "Message displayed to the user after failing to upload the local whitelist."
+        "description": "Message displayed to the user after failing to upload local data."
     }
 }
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -513,24 +513,36 @@
             }
         }
     },
+    "sync_intro": {
+        "message": "Cloud syncing:<ul><li>Requires Firefox/Chrome Sync</li><li>Download combines the lists of sites where your Badger is disabled</li><li>Upload overwrites the list saved in the cloud</li></ul>",
+        "description": "A brief explanation of how syncing works. Shown above the upload/download disabled sites buttons under the Manage Data options page tab."
+    },
     "upload": {
         "message": "Upload",
-        "description": ""
-    },
-    "merge": {
-        "message": "Merge",
-        "description": ""
+        "description": "Button label for Sync Upload on Manage Data options page tab."
     },
     "upload_whitelist": {
-        "message": "Upload disabled sites",
-        "description": ""
+        "message": "Upload disabled sites to cloud",
+        "description": "Button title for Sync Upload on Manage Data options page tab."
     },
     "download_whitelist": {
-        "message": "Download disabled sites",
-        "description": ""
+        "message": "Download disabled sites from cloud",
+        "description": "Button title for Sync Download on Manage Data options page tab."
     },
-    "sync_whitelist": {
-        "message": "Sync disabled sites to cloud",
-        "description": ""
+    "download_whitelist_success": {
+        "message": "Disabled sites list downloaded from cloud and merged into local list.",
+        "description": "Message displayed to the user after successfully downloading the synced whitelist."
+    },
+    "download_whitelist_failure": {
+        "message": "Disabled sites list could not be downloaded from cloud.  (You will get this message if you have not uploaded any list to the cloud.)",
+        "description": "Message displayed to the user after failing to download the synced whitelist."
+    },
+    "upload_whitelist_success": {
+        "message": "Disabled sites list uploaded to cloud.",
+        "description": "Message displayed to the user after successfully uploading the local whitelist."
+    },
+    "upload_whitelist_failure": {
+        "message": "Disabled sites list could not be uploaded to cloud.",
+        "description": "Message displayed to the user after failing to upload the local whitelist."
     }
 }
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -512,5 +512,25 @@
                 "example": "Facebook Like"
             }
         }
+    },
+    "upload": {
+        "message": "Upload",
+        "description": ""
+    },
+    "merge": {
+        "message": "Merge",
+        "description": ""
+    },
+    "upload_whitelist": {
+        "message": "Upload disabled sites",
+        "description": ""
+    },
+    "download_whitelist": {
+        "message": "Download disabled sites",
+        "description": ""
+    },
+    "sync_whitelist": {
+        "message": "Sync disabled sites to cloud",
+        "description": ""
     }
 }
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -530,19 +530,23 @@
         "description": "Button title for downloading data from the cloud on Manage Data options page tab."
     },
     "download_cloud_success": {
-        "message": "Disabled sites list downloaded from cloud and merged into local list.",
+        "message": "Cloud data imported successfully.",
         "description": "Message displayed to the user after successfully merging cloud data."
     },
     "download_cloud_failure": {
-        "message": "Disabled sites list could not be downloaded from cloud.  (You will get this message if you have not uploaded any list to the cloud.)",
+        "message": "Cloud data could not be downloaded.",
         "description": "Message displayed to the user after failing to download cloud data."
     },
     "upload_cloud_success": {
-        "message": "Disabled sites list uploaded to cloud.",
+        "message": "Cloud data uploaded successfully.",
         "description": "Message displayed to the user after successfully uploading local data."
     },
     "upload_cloud_failure": {
-        "message": "Disabled sites list could not be uploaded to cloud.",
+        "message": "Cloud data could not be uploaded.",
         "description": "Message displayed to the user after failing to upload local data."
+    },
+    "download_cloud_no_data": {
+        "message": "No cloud data to download.",
+        "description": "'Error' message when there is on cloud data to download."
     }
 }
--- a/src/js/options.js
+++ b/src/js/options.js
@@ -52,8 +52,8 @@ function loadOptions() {
   // Add event listeners
   $("#whitelistForm").on("submit", addWhitelistDomain);
   $("#removeWhitelist").on("click", removeWhitelistDomain);
-  $("#uploadWhitelist").on("click", uploadWhitelist);
-  $("#downloadWhitelist").on("click", downloadWhitelist);
+  $("#cloud-upload").on("click", uploadWhitelist);
+  $("#cloud-download").on("click", downloadWhitelist);
   $('#importTrackerButton').on("click", loadFileChooser);
   $('#importTrackers').on("change", importTrackerList);
   $('#exportTrackers').on("click", exportUserData);
@@ -101,8 +101,8 @@ function loadOptions() {
   $(".refreshButton").button("option", "icons", {primary: "ui-icon-refresh"});
   $(".addButton").button("option", "icons", {primary: "ui-icon-plus"});
   $(".removeButton").button("option", "icons", {primary: "ui-icon-minus"});
-  $("#uploadWhitelist").button("option", "icons", {primary: "ui-icon-arrowreturnthick-1-n"});
-  $("#downloadWhitelist").button("option", "icons", {primary: "ui-icon-arrowreturnthick-1-s"});
+  $("#cloud-upload").button("option", "icons", {primary: "ui-icon-arrowreturnthick-1-n"});
+  $("#cloud-download").button("option", "icons", {primary: "ui-icon-arrowreturnthick-1-s"});
   $(".importButton").button("option", "icons", {primary: "ui-icon-plus"});
   $("#exportTrackers").button("option", "icons", {primary: "ui-icon-extlink"});
   $("#resetData").button("option", "icons", {primary: "ui-icon-arrowrefresh-1-w"});
@@ -232,10 +232,10 @@ function downloadWhitelist() {
   chrome.runtime.sendMessage({type: "downloadWhitelist"},
     function (success) {
       if (success) {
-        alert(i18n.getMessage("download_whitelist_success"));
+        alert(i18n.getMessage("download_cloud_success"));
         reloadWhitelist();
       } else {
-        alert(i18n.getMessage("download_whitelist_failure"));
+        alert(i18n.getMessage("download_cloud_failure"));
       }
     }
   );
@@ -245,9 +245,9 @@ function uploadWhitelist() {
   chrome.runtime.sendMessage({type: "uploadWhitelist"},
     function (success) {
       if (success) {
-        alert(i18n.getMessage("upload_whitelist_success"));
+        alert(i18n.getMessage("upload_cloud_success"));
       } else {
-        alert(i18n.getMessage("upload_whitelist_failure"));
+        alert(i18n.getMessage("upload_cloud_failure"));
       }
     }
   );
--- a/src/js/options.js
+++ b/src/js/options.js
@@ -52,8 +52,8 @@ function loadOptions() {
   // Add event listeners
   $("#whitelistForm").on("submit", addWhitelistDomain);
   $("#removeWhitelist").on("click", removeWhitelistDomain);
-  $("#cloud-upload").on("click", uploadWhitelist);
-  $("#cloud-download").on("click", downloadWhitelist);
+  $("#cloud-upload").on("click", uploadCloud);
+  $("#cloud-download").on("click", downloadCloud);
   $('#importTrackerButton').on("click", loadFileChooser);
   $('#importTrackers').on("change", importTrackerList);
   $('#exportTrackers').on("click", exportUserData);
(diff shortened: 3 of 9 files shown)


Jsnhlbr5:
Do we want syncing of the disabled sites list to be a setting; if so, do we want the default to be on or off?

ghostwords:
I think it's fine to leave syncing preferences to the browser. If the user has syncing disabled, `chrome.storage.sync` will work the same as `chrome.storage.local`, right?

You could see what other popular open source extensions that use `chrome.storage.sync` do.

Jsnhlbr5:
Would it be reasonable to sync the entire settings_map (10 booleans and a number (mostly the checkboxes under general settings), in addition to the disabled list), rather than splitting out the disabled list?  I think doing so would be significantly easier and still fit within the sync quota.  We still would not sync the action_map or snitch_map, which hold the learning stuff.

ghostwords:
Perhaps but I suggest holding off and supporting only the disabled sites list to begin with. It seems like it would be simpler to do just the site domains for now. For instance, not all of `settings_map` is general settings; at least one item (`migrationLevel`) is closely linked to action/snitch maps.

Jsnhlbr5:
From a code standpoint, I think supporting only the disabled list is more complicated, but the `migrationLevel` makes it necessary (for now anyway).

Jsnhlbr5:
I believe that I have implemented this functionality now.  As a side effect, we are now storing the disabled sites list twice; once in the settings_map (locally) and once by itself in synced storage.  I did some manual testing, and to the best of my knowledge these changes work as intended, but I have not created any new tests (not sure how automated tests would even work for this).  I also ran into some difficulties getting ESLint installed/working (I am on Windows 10), so I didn't run that report either.

Jsnhlbr5:
Personally, I'm not too worried about exceeding `QUOTA_BYTES`, as that would mean the list was ~50k characters, but the loss/overwrite concern is definitely valid.  To copy uBlock, I see 3 new functions in options.js supporting 3 new buttons on the whitelist tab of settings; upload, download-overwrite, and download-merge.  I would also remove the code that I added to automatically sync.  Does this fit your expectations?

Jsnhlbr5:
I believe manual sync is now complete.  (I am open to feedback on the icons/labels I used, I just threw in something I thought was reasonable.)

src/js/webrequest.js
```diff
@@ -738,6 +738,43 @@ function dispatcher(request, sender, sendResponse) {
...
+        //TODO display a message to the user?
+        console.log("No whitelist in sync storage.");
+      }
+    });
+
+  } else if (request.type == "uploadWhitelist") {
+    var obj = {};
+    obj.disabledSites = badger.getSettings().getItem("disabledSites");
+    chrome.storage.sync.set(obj, function () {
+      if (chrome.runtime.lastError) {
```
ghostwords: We don't need to do the `criticalError` stuff for `storage.sync` saving since it will always be kicked off by the user and we will use alerts on the options page to provide feedback.

The `criticalError` system was added to handle reporting issues with updating Badger data to the user (so that they can in turn let us know about the problem): #1848. Badger data can get updated by all sorts of things; usually the updating isn't connected to the user interacting with Privacy Badger at all. In our case though we know the user just clicked a particular button on the options page.

src/skin/options.html
```diff
@@ -183,6 +183,18 @@ <h1><span class="i18n_options_title"></span></h1>
...
 
+    <div class="btn-silo">
+      <h2 id="sync-heading"><span class="i18n_sync_whitelist"></span></h2>
+      <div id="upload">
+        <h3><span class="i18n_upload_whitelist"></span></h3>
+        <button id="uploadWhitelist"><span class="i18n_upload"></span></button>
+      </div>
+      <div id="download">
+        <h3><span class="i18n_download_whitelist"></span></h3>
+        <button id="downloadWhitelist"><span class="i18n_download"></span></button>
```
ghostwords: Do we need both "download" and "merge"? I think just "merge" should be enough. We may not want to call it "merge" either ... Importing user data also merges disabled site lists. So, get rid of "download" and rename "merge" to "download"?

src/js/options.js
```diff
@@ -224,11 +230,37 @@ function removeAllData() {
   }
 }
 
+//TODO Do we want confirmations before or success/failure messages after for these?
+function downloadWhitelist() {
+  chrome.runtime.sendMessage({type: "downloadWhitelist"}, () => {
+    //Set a timeout because this needs to happen just a tiny bit later than it was without.
+    setTimeout(function() {
```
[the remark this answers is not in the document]
Jsnhlbr5: We need to refresh the whitelist so that it correctly displays the updated values, and in my testing, the refresh seemed to be happening before the data actually got set (the whitelist wouldn't be refreshed after downloading, but manually calling this function from the console on the settings page would update it).

src/js/options.js
```diff
@@ -224,11 +230,37 @@ function removeAllData() {
...
+  chrome.runtime.sendMessage({type: "uploadWhitelist"});
+}
+
 /**
  * Export the user's data, including their list of trackers from
  * action_map and snitch_map, along with their settings.
  * List will be in JSON format that can be edited and reimported
  * in another instance of Privacy Badger.
+ * 
+ * TODO This assumes that the local data is up to date with the
```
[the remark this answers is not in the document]
Jsnhlbr5: Oh, yeah, that was left over from the initial plan of automatic syncing.

src/_locales/en_US/messages.json
```diff
@@ -512,5 +512,25 @@
                 "example": "Facebook Like"
             }
         }
+    },
+    "upload": {
+        "message": "Upload",
+        "description": ""
```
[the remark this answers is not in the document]
Jsnhlbr5: To my quick perusal of the list, the other buttons/labels are missing descriptions as well; should I update those while I'm in there?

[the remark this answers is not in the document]
ghostwords: No need, we can fix them later. Some buttons/labels do have descriptions though, and we should provide them going forward.

src/skin/options.html
```diff
@@ -183,6 +183,18 @@ <h1><span class="i18n_options_title"></span></h1>
 
   <div id="tab-manage-data">
 
+    <div class="btn-silo">
```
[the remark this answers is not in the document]
Jsnhlbr5: Moved down, explanation added (no links, though)

src/js/options.js
```diff
@@ -224,11 +230,37 @@ function removeAllData() {
   }
 }
 
+//TODO Do we want confirmations before or success/failure messages after for these?
```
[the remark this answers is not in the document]
Jsnhlbr5: success/failure alerts added to both upload and download

src/js/webrequest.js
```diff
@@ -738,6 +738,43 @@ function dispatcher(request, sender, sendResponse) {
...
     sendResponse();
 
+  } else if (request.type == "downloadWhitelist") {
+    chrome.storage.sync.get("disabledSites", function (store) {
+      if (store.hasOwnProperty("disabledSites")) {
+        badger.getSettings().setItem("disabledSites", store.disabledSites);
+        sendResponse();
+      } else {
+        //TODO display a message to the user?
+        console.log("No whitelist in sync storage.");
```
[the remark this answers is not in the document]
Jsnhlbr5: done

src/js/webrequest.js
```diff
@@ -738,6 +738,43 @@ function dispatcher(request, sender, sendResponse) {
...
+
+  } else if (request.type == "mergeWhitelist") {
+    chrome.storage.sync.get("disabledSites", function (store) {
+      if (store.hasOwnProperty("disabledSites")) {
+        var settings = badger.getSettings();
+        var whitelist = _.union(settings.getItem("disabledSites"), store.disabledSites);
+        settings.setItem("disabledSites", whitelist);
+        sendResponse();
+      } else {
+        //TODO display a message to the user?
```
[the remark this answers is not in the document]
Jsnhlbr5: done

src/skin/options.html
Jsnhlbr5: removed merge, download merges lists

src/_locales/en_US/messages.json
```diff
@@ -512,5 +512,37 @@
...
+    },
+    "sync_intro": {
+        "message": "Cloud syncing:<ul><li>Requires Firefox/Chrome Sync</li><li>Download combines the lists of sites where your Badger is disabled</li><li>Upload overwrites the list saved in the cloud</li></ul>",
+        "description": "A brief explanation of how syncing works. Shown above the upload/download disabled sites buttons under the Manage Data options page tab."
+    },
+    "upload": {
+        "message": "Upload",
+        "description": "Button label for Sync Upload on Manage Data options page tab."
+    },
+    "upload_whitelist": {
```
ghostwords: We should rename all of these message keys to say "cloud" instead of "whitelist" since we may choose to upload more than just the list of disabled sites in the future.

src/_locales/en_US/messages.json
```diff
@@ -512,5 +512,37 @@
...
+        "message": "Upload disabled sites to cloud",
+        "description": "Button title for Sync Upload on Manage Data options page tab."
+    },
+    "download_whitelist": {
+        "message": "Download disabled sites from cloud",
+        "description": "Button title for Sync Download on Manage Data options page tab."
+    },
+    "download_whitelist_success": {
+        "message": "Disabled sites list downloaded from cloud and merged into local list.",
+        "description": "Message displayed to the user after successfully downloading the synced whitelist."
```
ghostwords: No need to mention what is being synced inside the descriptions (we may sync more than just disabled sites in the future).

"... after successfully merging cloud data."

src/_locales/en_US/messages.json
```diff
@@ -512,5 +512,37 @@
...
+    },
+    "download_whitelist_failure": {
+        "message": "Disabled sites list could not be downloaded from cloud.  (You will get this message if you have not uploaded any list to the cloud.)",
+        "description": "Message displayed to the user after failing to download the synced whitelist."
```
ghostwords: "... failing to download cloud data."

src/_locales/en_US/messages.json
```diff
@@ -512,5 +512,37 @@
...
+    },
+    "upload_whitelist_success": {
+        "message": "Disabled sites list uploaded to cloud.",
+        "description": "Message displayed to the user after successfully uploading the local whitelist."
```
ghostwords: "... uploading local data."

src/_locales/en_US/messages.json
```diff
@@ -512,5 +512,37 @@
...
+    },
+    "upload_whitelist_failure": {
+        "message": "Disabled sites list could not be uploaded to cloud.",
+        "description": "Message displayed to the user after failing to upload the local whitelist."
```
ghostwords: "... upload local data."

.gitignore
```diff
@@ -54,3 +54,7 @@ web-ext-artifacts
 pkg
 release-utils/config.sh
 release-utils/xpi
+
+#Eclipse project stuff
```
ghostwords: Could we move these exclusions to somewhere above Privacy Badger-specific exclusions (`release-utils`, etc.)?

src/js/webrequest.js
```diff
@@ -738,6 +738,31 @@ function dispatcher(request, sender, sendResponse) {
     badger.storage.revertUserAction(request.origin);
     sendResponse();
 
+  } else if (request.type == "downloadWhitelist") {
+    chrome.storage.sync.get("disabledSites", function (store) {
+      if (store.hasOwnProperty("disabledSites")) {
+        var settings = badger.getSettings();
```
ghostwords: Could we use `let` or `const` (when appropriate) instead of `var`? I know we have `var` all over the place right now; no need to fix existing usage, but let's switch to `let`/`const` going forward.

src/_locales/en_US/messages.json
```diff
@@ -512,5 +512,37 @@
...
             }
         }
+    },
+    "sync_intro": {
+        "message": "Cloud syncing:<ul><li>Requires Firefox/Chrome Sync</li><li>Download combines the lists of sites where your Badger is disabled</li><li>Upload overwrites the list saved in the cloud</li></ul>",
+        "description": "A brief explanation of how syncing works. Shown above the upload/download disabled sites buttons under the Manage Data options page tab."
+    },
+    "upload": {
+        "message": "Upload",
+        "description": "Button label for Sync Upload on Manage Data options page tab."
```
ghostwords: "... for uploading data to the cloud ..."

src/_locales/en_US/messages.json
```diff
@@ -512,5 +512,37 @@
                 "example": "Facebook Like"
             }
         }
+    },
+    "sync_intro": {
+        "message": "Cloud syncing:<ul><li>Requires Firefox/Chrome Sync</li><li>Download combines the lists of sites where your Badger is disabled</li><li>Upload overwrites the list saved in the cloud</li></ul>",
```
ghostwords: Just "Cloud sync:" instead of "Cloud syncing:"?

ghostwords: The "upload" bullet point should probably come before the "download" bullet point, to match button order.

ghostwords: "Upload overwrites ~~the list saved in the~~any existing data in the cloud"? To avoid getting tied to what is specifically saved.

src/skin/options.html
```diff
@@ -194,6 +194,17 @@ <h3><span class="i18n_export_user_data"></span></h3>
         <button id="exportTrackers"><span class="i18n_download"></span></button>
       </div>
     </div>
+    <div class="btn-silo">
+      <p class="i18n_sync_intro"></p>
```
ghostwords: Could we fix margin/padding around this paragraph to be a bit more balanced? It feels like there is too much space at the bottom and too little at the top.

src/skin/options.html
```diff
@@ -194,6 +194,17 @@ <h3><span class="i18n_export_user_data"></span></h3>
...
+      <div id="upload">
+        <h3><span class="i18n_upload_whitelist"></span></h3>
+        <button id="uploadWhitelist"><span class="i18n_upload"></span></button>
```
ghostwords: Should rename the IDs too ... "uploadWhitelist" to "cloud-upload"? (CSS names should be dash-separated and lowercase going forward.)

src/js/webrequest.js
```diff
@@ -738,6 +738,31 @@ function dispatcher(request, sender, sendResponse) {
...
+        var whitelist = _.union(settings.getItem("disabledSites"), store.disabledSites);
+        settings.setItem("disabledSites", whitelist);
+        sendResponse(true);
+      } else {
+        sendResponse(false);
+      }
+    });
+    return true; //Let internal asychronus call continue
```
ghostwords: typo: "asynchronous"

src/js/webrequest.js
```diff
@@ -738,6 +738,31 @@ function dispatcher(request, sender, sendResponse) {
...
+    var obj = {};
+    obj.disabledSites = badger.getSettings().getItem("disabledSites");
+    chrome.storage.sync.set(obj, function () {
+      if (chrome.runtime.lastError) {
+        sendResponse(false);
+      } else {
+        sendResponse(true);
+      }
+    });
+    return true; //Let internal asychronus call continue
```
ghostwords: typo

src/js/webrequest.js
ghostwords: Should we reuse previously-used comment text, "indicate this is an async response to chrome.runtime.onMessage"?

src/_locales/en_US/messages.json
ghostwords: Should we change the order of the Import/Download user data buttons above to match the order of the Upload/Download buttons? So it would be "Download" (export) and "Import" (import) user data, and then "Upload" (export) and "Download" (import) cloud data.

src/js/options.js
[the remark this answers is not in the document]
Jsnhlbr5: I now believe this was actually caused by the way that the onMessage event handler executes, which was cutting off the callback from the `chrome.storage.sync.get` function.  (The solution to that is to return true from the event handler; this was found when debugging the success/failure alert always displaying failure.)

src/_locales/en_US/messages.json
Jsnhlbr5: I would be inclined to change the order of the new buttons, not the old, but will defer to your judgement.

ghostwords: Just seems like importing should line up with importing. Not sure whether it's better to have import or export first. Maybe export? I think it's fine to reorder existing UI.

src/_locales/en_US/messages.json
Jsnhlbr5: done

src/_locales/en_US/messages.json
Jsnhlbr5: done

src/_locales/en_US/messages.json
Jsnhlbr5: done

src/_locales/en_US/messages.json
Jsnhlbr5: done

src/_locales/en_US/messages.json
Jsnhlbr5: done

src/_locales/en_US/messages.json
Jsnhlbr5: While avoiding getting tied to what is saved is _fine_ on the upload bullet, I do think the download bullet should describe current behavior.

ghostwords: Agreed.

src/skin/options.html
Jsnhlbr5: I added a bit at the top and took off all the margin at the bottom, but the padding/margin on the div/h3 below it is responsible for ~2/3 of that space.  I'd prefer not to make those special for this row.

src/js/webrequest.js
Jsnhlbr5: I had not noticed that comment before, I have changed them.

ghostwords:
@bcyphers @andresbase I'm not sure about Export/Import, Download/Upload button order and button labels. We may also want to tweak the explanation. What do you guys think? We could merge this soon and keep making tweaks as part of #2145.

src/_locales/en_US/messages.json
```diff
@@ -512,5 +512,37 @@
...
+    "download_cloud": {
+        "message": "Download disabled sites from cloud",
+        "description": "Button title for downloading data from the cloud on Manage Data options page tab."
+    },
+    "download_cloud_success": {
+        "message": "Disabled sites list downloaded from cloud and merged into local list.",
+        "description": "Message displayed to the user after successfully merging cloud data."
+    },
+    "download_cloud_failure": {
+        "message": "Disabled sites list could not be downloaded from cloud.  (You will get this message if you have not uploaded any list to the cloud.)",
```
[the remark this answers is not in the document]
Jsnhlbr5: Do we want to append the error message? (Would it be localized?  How likely is it to be too technical for the average user?)  We could instead have a standard message saying to check the error console for details.

src/_locales/en_US/messages.json
```diff
@@ -512,5 +512,37 @@
...
+        "description": "Message displayed to the user after failing to download cloud data."
+    },
+    "upload_cloud_success": {
+        "message": "Disabled sites list uploaded to cloud.",
+        "description": "Message displayed to the user after successfully uploading local data."
+    },
+    "upload_cloud_failure": {
+        "message": "Disabled sites list could not be uploaded to cloud.",
```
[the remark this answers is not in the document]
Jsnhlbr5: See above comment/question.

src/_locales/en_US/messages.json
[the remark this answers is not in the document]
ghostwords: We could append it. It might be localized, but it's fine either way. It's probably too technical but then we do have clear error text above (something like "Failed to download cloud data"), and displaying the exact error will help us figure out the issue if users start reporting these failures to us. On the other hand, `console.error`-printing them works too since we could just ask users to open the options page console. So ... either is fine, but we should do something.

andresbase:
For the time being the order of buttons and text make sense. For reference this is what it currently looks like:
screen shot 2018-08-30 at 13 51 30

Test in Chrome worked as expected. Will try later with FF.

src/_locales/en_US/messages.json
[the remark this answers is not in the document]
Jsnhlbr5: I figured that writing to `console.error` was not a bad idea either way, so that's what I have in there right now.

src/_locales/en_US/messages.json
```diff
@@ -508,5 +508,41 @@
                 "example": "Facebook Like"
             }
         }
+    },
+    "sync_intro": {
+        "message": "Cloud sync:<ul><li>Requires Firefox/Chrome Sync</li><li>Upload overwrites any existing data in the cloud</li><li>Download combines the lists of sites where your Badger is disabled</li></ul>",
```
bcyphers: Maybe change the second bullet to read
> - Upload overwrites any existing **Privacy Badger** data in the cloud

Jsnhlbr5: good clarification

src/_locales/en_US/messages.json
```diff
@@ -508,5 +508,41 @@
...
+    "upload": {
+        "message": "Upload",
+        "description": "Button label for uploading data to the cloud on Manage Data options page tab."
+    },
+    "upload_cloud": {
+        "message": "Upload disabled sites to cloud",
+        "description": "Button title for uploading data to the cloud on Manage Data options page tab."
+    },
+    "download_cloud": {
+        "message": "Download disabled sites from cloud",
```
ghostwords: "Import ...", for consistency with file export/import labels.

src/_locales/en_US/messages.json
```diff
@@ -508,5 +508,41 @@
...
+    "sync_intro": {
+        "message": "Cloud sync:<ul><li>Requires Firefox/Chrome Sync</li><li>Upload overwrites any existing Privacy Badger data in the cloud</li><li>Download combines the lists of sites where your Badger is disabled</li></ul>",
+        "description": "A brief explanation of how syncing works. Shown above the upload/download cloud data buttons under the Manage Data options page tab."
+    },
+    "upload": {
+        "message": "Upload",
+        "description": "Button label for uploading data to the cloud on Manage Data options page tab."
+    },
+    "upload_cloud": {
+        "message": "Upload disabled sites to cloud",
```
ghostwords: "Export ...", for consistency with file export/import labels.

src/skin/options-layout.css
```diff
@@ -60,6 +44,13 @@ button
...
+  width: 350px;
+  float: left;
+}
+
+.btn-silo p {
+  padding-top: 10px;
+  margin-bottom: 0;
 }
 
 .btn-silo p {
```
ghostwords: Remove this one or the one above (got duplicated by merge from master).

src/skin/options-layout.css
```diff
@@ -60,6 +44,13 @@ button
 
 .btn-silo div {
   padding-top: 10px;
```
ghostwords: Spacing tweak?
```diff
 .btn-silo + .btn-silo {
   border-top: 1px solid #d3d3d3;
-  margin-top: 10px;
+  margin-top: 20px;
+  padding-top: 10px;
 }
 
 .btn-silo div {
-  padding-top: 10px;
   width: 350px;
   float: left;
 }
```

Jsnhlbr5: I don't think we want to increase that top margin (that just puts more space between the buttons and the separator line below, and that spacing I think is fine), but moving the padding looks good.

andresbase:
LGTM

src/_locales/en_US/messages.json
```diff
@@ -508,5 +508,41 @@
...
+    "sync_intro": {
+        "message": "Cloud sync:<ul><li>Requires Firefox/Chrome Sync</li><li>Upload overwrites any existing Privacy Badger data in the cloud</li><li>Download combines the lists of sites where your Badger is disabled</li></ul>",
+        "description": "A brief explanation of how syncing works. Shown above the upload/download cloud data buttons under the Manage Data options page tab."
+    },
+    "upload": {
+        "message": "Upload",
+        "description": "Button label for uploading data to the cloud on Manage Data options page tab."
+    },
+    "upload_cloud": {
+        "message": "Export disabled sites to cloud",
```
bcyphers: I think changing the phrasing here from "import/export" to "upload/download" would be more consistent and easier to understand.

src/_locales/en_US/messages.json
```diff
@@ -508,5 +508,41 @@
...
+        "description": "Button title for uploading data to the cloud on Manage Data options page tab."
+    },
+    "download_cloud": {
+        "message": "Import disabled sites from cloud",
```
bcyphers: See above -- "Download disabled sites from cloud"

src/_locales/en_US/messages.json
```diff
@@ -508,5 +508,41 @@
...
+        "description": "Button title for downloading data from the cloud on Manage Data options page tab."
+    },
+    "download_cloud_success": {
+        "message": "Cloud data imported successfully.",
```
bcyphers: imported -> downloaded?

src/_locales/en_US/messages.json
```diff
@@ -508,5 +508,41 @@
...
+    },
+    "upload_cloud_success": {
+        "message": "Cloud data uploaded successfully.",
+        "description": "Message displayed to the user after successfully uploading local data."
+    },
+    "upload_cloud_failure": {
+        "message": "Cloud data could not be uploaded.",
+        "description": "Message displayed to the user after failing to upload local data."
+    },
+    "download_cloud_no_data": {
```
bcyphers: I never got this message to trigger -- before I uploaded anything, the download button gave the "download_cloud_failure" message, and after uploading empty data, it gave the "success" message. See comment in `options.js`.

src/js/options.js
```diff
@@ -224,6 +228,33 @@ function removeAllData() {
...
 }
 
+function downloadCloud() {
+  chrome.runtime.sendMessage({type: "downloadCloud"},
+    function (status) {
+      if (status.success) {
+        alert(i18n.getMessage("download_cloud_success"));
+        reloadWhitelist();
+      } else {
+        alert(i18n.getMessage("download_cloud_failure"));
```
bcyphers: The message handler in `webrequest.js` returns `{success: false, message: message}`, but the code here alerts with the "failure" message no matter what. This should check `status.message` and decide whether to alert with the failure message or the "no data" message.

src/_locales/en_US/messages.json
Jsnhlbr5: @ghostwords suggested import/export to be consistent with the other pair of user data buttons.

src/_locales/en_US/messages.json
Jsnhlbr5: To get this message, something has to go wrong writing to the synced storage (network connectivity loss during write might do it, as would going over quota which is 100kB total, 8kB per item [with the list of disabled sites being 1 item] see https://developer.chrome.com/apps/storage#properties)

src/js/options.js
Jsnhlbr5: The message was mostly there for error debugging, but this is a reasonable enhancement.

src/_locales/en_US/messages.json
Jsnhlbr5: This user data file I manipulated to have too long of a disabled sites list, and produces the error condition `QUOTA_BYTES_PER_ITEM quota exceeded`.
[PrivacyBadger_user_data-long-DL.zip](https://github.com/EFForg/privacybadger/files/2350817/PrivacyBadger_user_data-long-DL.zip)

src/js/options.js
bcyphers: Ah, ok, makes sense. I do think it would be useful for the user to differentiate between the failure modes.

src/_locales/en_US/messages.json
bcyphers: Oh ok! didn't see that conversation. I'll defer to his judgment here.

src/js/options.js
Jsnhlbr5: After making this change, I'm not sure what would make the failure message (as opposed to the no data message) appear; write conditions can be manipulated to exceed quotas and produce errors, but read operations have no such quotas.  Presumably, loss of network connectivity during the transfer would do it, but I'm not sure how to reliably replicate that for such a small download ( <8kB by definition).  No network connectivity appears to fall back on local cache.

ghostwords:
@Jsnhlbr5 Thanks again!

PR #2154, opened 2018-09-02
Keep scroll position when toggling slider for a domain

### LINKED ISSUE #1581: Toggling slider for a domain on options page resets scroll

troy-lamerton:
Fixes #1581

### DIFF
--- a/src/js/options.js
+++ b/src/js/options.js
@@ -465,6 +465,7 @@ function revertDomainControl(e) {
  * Displays list of all tracking domains along with toggle controls.
  */
 function reloadTrackingDomainsTab() {
+  var blockedResourcesScrollPosition = $("#blockedResourcesInner").scrollTop();
   refreshOriginCache();
 
   // Check to see if any tracking domains have been found before continuing.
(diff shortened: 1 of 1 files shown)


troy-lamerton:
Is the trackers list being refreshed so that if the toggled domain is now filtered, it is removed?
Such as here, after a slider is moved: https://github.com/EFForg/privacybadger/blob/master/src/js/options.js#L595

Then I would simply check if the toggled domain or 'reverted to privacy badger' domain is filtered out now, and remove it from the DOM if this is the case.

bcyphers:
Good question! I think it's also to propagate changes to subdomains, if necessary -- for example:

toggle-domain-propagate

If we can handle both of those changes without refreshing the entire page, that might be worth doing.

ghostwords:
Replaced by #2629.

PR #2287, opened 2019-02-01
Updated and standardised social logos

### LINKED ISSUE #1857: Google+ icon is outdated

RemakingEden:
Fixes #1857.

I have updated and standardised all logos. I have taken a slightly different design to the past which should help the buttons survive longer without looking outdated. The changes I have made are:

1. Removed logos (These get updated and changed often and are different sizes)
2. Put the name of the action on each button
3. Made the button the same colour as the company in question
4. Standardised the button size. There is 2 sizes a larger one and a smaller one depending on how much text was needed.

It seems there may be an issue with the Google plus button however it is an issue with it being fetched and is code change which will take me a little bit of time to understand and fix. However a new updated logo is ready when this is fixed.


ghostwords:
Thank you for the PR!

New vs. old (ignore the broken Pinterest image, that's an existing bug):
52072471-05b87a00-2553-11e9-92f8-e80ea1a56752

RemakingEden:
@ghostwords, my pleasure.

I’m happy to make those changes I’ll just wait for @bcyphers and @andresbase opinion on them before making the changes.

I’m happy to make them SVG if it can be figured out. Do you have any idea how to make the size adapt? Does it need to be hard coded?

andresbase:
I think some of the confusion comes from not being clear this is blocked. In the old style since it looks like the Badger icon is on top of the "sharing" icon it sort of looks like I did something. 

The new icons look better but I'm not sure they communicate "I blocked this, you need to click". Is there a way that can be more evident? 

For Facebook the problem is that not everyone will understand that "Like" means Facebook. We shouldn't rely on colors for messaging since color blind people won't be able to tell the difference.

This is looking great though and it would be nice to modernize and simplify our iconography, thanks!

andresbase:
A tool that can help is doing a contrast test https://webaim.org/resources/contrastchecker/

RemakingEden:
@andresbase I see, good point on the colour blind. I hadn't even considered!

As for the blocking I understand. I cannot think of anyway around it without going the current way with the logos. It's not bad that way it's just hard to update as its hard to get a good quality logo without taking it from a low quality jpeg. Also it will look old probably within the next few months.

That being said, we could look at just using the company logo and no like thumb etc? That could probably hold the test of time for a year or two. What do you think?

RemakingEden:
@andresbase any idea about this? Are you happy with above?

andresbase:
@RemakingEden This seems to work better in several environments. I'm guessing the border cutting the badger logo wouldn't be seen.

The overhanging badger could as you mention cause some formatting issues. How about extending the button to include the badger inside? 

I'm thinking an other option would be to put the badger over the logo but not centered over each other. So you would see 60% or so of the social network logo creeping under the badger badge. Not sure how those would really look and if it's easy to get consistency across logos.

sta55en:
Hey folks, my 2c if I may :-)

I'd like to mention that most brands have acceptable usage guidelines for their logos and how their logos are used. Most of these brand guidelines prevent 3rd parties from overlaying logos with anything. Most of these rules also include guidelines about how close logos are allowed to be to other logos, etc. An example of this can be found here: https://en.facebookbrand.com/assets/f-logo/ I have not checked if these images are violating these guidelines directly, but we should be mindful of their existence.

In terms of having a non-rectangular button; there is quite a bit of risk that icons with an offset badger would not render correctly, or would be scaled down drastically to fit into the available space for the button. Since we do not strictly control the container that these icons appear inside of, I'd suggest that a square or rectangular icon is used.

andresbase:
Thanks @sta55en this looks really nice! I'll check with our design team.

ghostwords:
@sta55en Thanks for your input! Would you mind opening a new pull request where you replace the SVGs in [`/src/skin/socialwidgets/`](https://github.com/EFForg/privacybadger/tree/master/src/skin/socialwidgets) with updated versions?

sta55en:
@ghostwords Sure thing... As soon as I have a sec I'll get this in.

PR #2333, opened 2019-03-21
Add NYU domains to MDFP list

northgardner:


bcyphers:
Has this caused any issues for you? I browsed around the sites a bit and didn't notice Privacy Badger blocking content or ID'ing resources from among NYU sites as trackers.

(We'd prefer not to add MDFP entries unless they fix a demonstrable issue.)

northgardner:
Yes, we have had a couple demonstrable issues, that seem to happen because we
do front-end searches against Solr hosts on *.dlib.nyu.edu from websites
that are not in nyu.edu domain (such as site on  the nyupress.org domain.).

Jonathan

bcyphers:
Thanks. Looks like there are some lint issues: https://travis-ci.org/EFForg/privacybadger/jobs/509630785#L467. Would you mind fixing those? After that it's good to go.

northgardner:
Will do. Thanks.

PR #2444, opened 2019-08-28
Ignore separator characters when computing entropy for pixel cookie sharing

ablanathtanalba:
WIP: This doesn't resolve each point in #2436 , but it does touch on the initial discovery that separator characters aren't being correctly handled in `estimateMaxEntropy`. These proposed changes remove those separator characters from the input string. There's a basic unit test put in place here, though that will have to be properly merged in after the bigger set of test coverage is finally merged in from #2425  .


ablanathtanalba:
Closed to rename the branch --- now known as `test-char-classes-pixel-tracking` and open as a PR in #2452

PR #2463, opened 2019-09-17
Make it more clear PB is an EFF project

ablanathtanalba:
Although there isn't an issue already filed on GitHub for this, per internal discussions we've decided that the extension popup ought to be more clear that Privacy Badger is a project of the EFF. This accomplishes that, as well as some minor design changes that we agreed upon in discussion. Still a WIP as we narrow in on what's within the scope of this PR, as well as possible seeking outside perspective on design change.


src/skin/popup.html
```diff
@@ -121,6 +124,7 @@ <h1 id="share_title" class="i18n_share_title overlay_title"></h1>
     <button id="error" class="pbButton" target="blank"><span class="i18n_report_broken_site"></span></button>
     <button id="donate" class="pbButton" target="blank"><span class="i18n_donate_to_eff"></span></button>
 </div>
+<span id="version" class="i18n_version">
```
ghostwords: Malformed HTML.

src/skin/popup.html
```diff
@@ -90,14 +90,17 @@ <h1 id="share_title" class="i18n_share_title overlay_title"></h1>
...
+  <a id="share" title="i18n_popup_share_button" class="control-button grey-button tooltip" style="display:block" target="_blank"><img width="25" height="auto" src="/icons/black-share-icon.svg" alt="i18n_popup_share_button"></a>
+  <a id="help" title="i18n_popup_help_button" href='/skin/firstRun.html' class="control-button grey-button tooltip" style="display:block" target="_blank"><img width="25" src="/icons/black-help-icon.svg" alt="i18n_popup_help_button"></a>
   <div id='title'>
-    <img src="/icons/badger-48.png" srcset="/icons/badger-128.png 2x" width="48" alt=""> <h2><span class="i18n_name"></span></h2>
+    <img src="/icons/PrivacyBadger-noborder.svg" srcset="/icons/PrivacyBadger-noborder.svg 2x" width="48" alt="">
+    <div id='header-image-stack'>
+      <img id='header-red-eff-logo' src="./images/EFF-red.svg" srcset="./images/EFF-red.svg" width="48" alt=""> <br>
+      <img id='privacy-badger-text-logo' src="./images/PrivacyBadger-ChunkFive-Black.svg" srcset="./images/PrivacyBadger-ChunkFive-Black.svg"  width="180">
```
ghostwords: We should instead style the localized name of Privacy Badger (`<span class="i18n_name"></span>`) with CSS. For examples, see how we use custom fonts on the new user welcome page.

src/skin/popup.html
ghostwords: There is now a vertical scrollbar for the entire popup in Firefox. The scrollbar is an existing issue (https://github.com/EFForg/privacybadger/issues/1077#issuecomment-340041201), but not in the default locale. Ideally, we would find a solution that works for all locales, but a quick tweaks that restores the default locale to present (no scrollbar) state is OK too. (Forcing the scrollbar to be hidden is not an acceptable tweak ... we don't want to have any cases where Badger buttons are entirely inaccessible.)

ghostwords:
Once we're close to done here, we should see what fixes we need to make to RTL language hacks in https://github.com/EFForg/privacybadger/blob/2213448632d7a6a44c5af0bb5d1fa2c96a91f908/src/lib/i18n.js#L50-L66

src/skin/popup.html
ablanathtanalba: Good call -- I'm bringing this ask around to the design team to see if they've got this font handy. If so, I'll import it and do the changes to the header text with css instead of with an img tag

src/skin/popup.html
ablanathtanalba: Ah, I'm seeing the vertical scrollbar now. It seems to change with the new badger img. I'm tweaking some of the padding and margin values on the header and tracker portions of the popup to narrow it down to bare essentials without making it seem too cramped. There are some sizing issues with the newer badger icon (with the less beady eyes), once I get that sorted out with design I'll share screenshots here.

src/skin/popup.html
ghostwords: We already have regular Chunk in the [fonts folder](https://github.com/EFForg/privacybadger/tree/master/src/skin/fonts). As per [credits in our license file](https://github.com/EFForg/privacybadger/blob/2213448632d7a6a44c5af0bb5d1fa2c96a91f908/LICENSE#L90) you should be able to get the bold version from https://www.theleagueofmoveabletype.com/.

ablanathtanalba: Just got the bold version from design -- incorporating now, as well as some minor fixing on the new badger logo

ablanathtanalba:
Still waiting on the resized logo with the correct eye size, but here are some screenshots of my local branch's working state: 

Here is the popup with the updated message for a special page:
Screen Shot 2019-09-23 at 5 36 15 PM

And here is the popup on a typical page with trackers (also showing the hover change effect on one of the bottom buttons):
Screen Shot 2019-09-23 at 5 35 51 PM

ablanathtanalba:
Finally got the updated badger icon from design (with the regular sized eyes), here are some screenshots of the popup from the most recent commit. The issue with the scrollbar appearing over the entire popup has been fixed. Some minor spacing things have been tweaked as well -- including some space between the badger icon and the stack of logos in the header (a request from design).

The popup after initial download/on any special pages site.
Screen Shot 2019-09-30 at 10 08 48 AM


the popup page with trackers:
Screen Shot 2019-09-30 at 10 08 32 AM

src/icons/PrivacyBadger-noborder.svg
```diff
@@ -0,0 +1,21 @@
+<svg id="Group_928" data-name="Group 928" xmlns="http://www.w3.org/2000/svg" width="150.631" height="108.11" viewBox="0 0 150.631 108.11">
```
ghostwords: This file is no longer used, right?

src/skin/popup.html
```diff
@@ -90,14 +90,17 @@ <h1 id="share_title" class="i18n_share_title overlay_title"></h1>
   </div>
 
 <div id='privacyBadgerHeader'>
-  <a id="options" title="i18n_popup_options_button" href='/skin/options.html' class="control-button grey-button tooltip" style="display:block" target="_blank"><img src="/icons/options.svg" alt="i18n_popup_options_button"></a>
-  <a id="share" title="i18n_popup_share_button" class="control-button grey-button tooltip" style="display:block" target="_blank"><img width="25" height="auto" src="/icons/share.svg" alt="i18n_popup_share_button"></a>
-  <a id="help" title="i18n_popup_help_button" href='/skin/firstRun.html' class="control-button grey-button tooltip" style="display:block" target="_blank"><img src="/icons/help.svg" alt="i18n_popup_help_button"></a>
+  <a id="options" title="i18n_popup_options_button" href='/skin/options.html' class="control-button grey-button tooltip" style="display:block" target="_blank"><img width="25" src="/icons/black-options-icon.svg" alt="i18n_popup_options_button"></a>
```
ghostwords: Why don't we just replace the existing icon files with the new icon files? No need for new filenames, is there?

src/skin/popup.html
```diff
@@ -90,14 +90,17 @@ <h1 id="share_title" class="i18n_share_title overlay_title"></h1>
...
+  <a id="share" title="i18n_popup_share_button" class="control-button grey-button tooltip" style="display:block" target="_blank"><img width="25" height="auto" src="/icons/black-share-icon.svg" alt="i18n_popup_share_button"></a>
+  <a id="help" title="i18n_popup_help_button" href='/skin/firstRun.html' class="control-button grey-button tooltip" style="display:block" target="_blank"><img width="25" src="/icons/black-help-icon.svg" alt="i18n_popup_help_button"></a>
   <div id='title'>
-    <img src="/icons/badger-48.png" srcset="/icons/badger-128.png 2x" width="48" alt=""> <h2><span class="i18n_name"></span></h2>
+    <img src="./images/PB-logo-bw.svg" srcset="./images/PB-logo-bw.svg 2x" width="48" alt="privacy badger logo" id="badger-header-logo">
```
ghostwords: Do we actually still need the `srcset` property?

ghostwords: The `alt` property should stay blank as it's not important to tell screen readers that there is a logo here; the logo is just decorative. We could label the logo "Privacy Badger", but there is no need as we also have actual text that says "Privacy Badger" right next to the logo. Please see https://webaim.org/techniques/alttext/ for more information.

src/skin/popup.html
ghostwords: Still needs doing.

src/skin/popup.html
ghostwords: Closing tag and scrollbar in Firefox are still problems.

src/skin/popup.html
ablanathtanalba: I thought the `srcset` attribute was a sort of safety precaution, but I guess since we aren't doing any dynamic rendering of different assets based on screen size, I can take this one and others in this html file out

ghostwords: Is there a point to a `srcset` that specifies the same image file twice?

ghostwords: I don't think you should remove the other `srcset`s ...

ghostwords:
You probably have trouble reproducing because of how macOS handles scrollbars by default.

ghostwords:
>we should see what fixes we need to make to RTL language hacks

Should also check what impact if any this will have on #1077 and #1113.

src/skin/popup.html
ablanathtanalba: Thought I fixed that in a previous commit, it's been fixed in the most recent one

src/icons/PrivacyBadger-noborder.svg
ablanathtanalba: Good call, this has been removed.

src/skin/popup.html
ablanathtanalba: Ah, I meant the previous one that I added. It's been removed

ghostwords:
Here is what the popup looks like for me after the most recent commit in Firefox on `nytimes.com`:

Screenshot from 2019-10-02 17:49:07

Note the scrollbar and the extra space in the header.

ablanathtanalba:
Thank you for the screenshot! I'm switching to a linux environment for this last task so that I can accurately recreate this scrollbar problem

ablanathtanalba:
I'm still unable to recreate the scrollbar error, in both a linux or a macos environment (and each tested with different resolution settings just in case). For reference, here's a screenshot of visiting nytimes in my ubuntu VM, running a fresh badger instance off this most recent commit:

Screen Shot 2019-10-15 at 5 38 51 PM

I suspect that with the merge of #2479 into master, this will no longer be an issue. I'll circle back to confirm after a rebase and some manual testing

src/skin/images/PrivacyBadger-ChunkFive-Black.svg
```diff
@@ -0,0 +1,17 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="448.686" height="58.966" viewBox="0 0 448.686 58.966">
```
ghostwords: No need to keep the "Privacy Badger" SVG, right?

ablanathtanalba:
As the request for comment goes through the design team's laundry cycles, here are recent updates:

workaround on RTL locales now looks like this. Just a proposed change, still waiting for formal review from design:
RTL_on_macos


Here is english LTR on the same MacOS environment:
LTR_on_macos

And here is the same locale on linux environment:
LTR_on_linux


I cut the margin space again out of the `blockedResourcesContainer` div in the popup so as to prevent the scrollbar from appearing in linux environments. It seems to me that the tradeoff is either we reduce that space, or the space between the now stacked elements in the header bar (which were pretty cramped before). Open to either, but we now only have 3px of vertical giving room in this current iteration before the scrollbar will reappear in linux environments.

PR #2481, opened 2019-10-16
Add Riot Games domains

v3n:
We've received reports from players visiting the page for our newly announced game, playruneterra.com, that Privacy Badger is flagging our beta sign-up flow as cross-site tracking.

This adds MDFP rules for Riot Games domains.


v3n:
Good idea :)

Added those.

PR #2485, opened 2019-10-22
Added The Vanguard Group domains to MDFP

Jsnhlbr5:
Currently, 'vgcontent.info' gets flagged as a potential tracker, and after visiting pages from the 3 other domains it gets blocked, thoroughly breaking the site.


Jsnhlbr5:
Yes, we should

PR #2551, opened 2020-02-12
Adding loveholidays.com and lovevacations.com domains to MDFP

dios-david:
Some of the localised sites (e.g. https://loveholidays.se , https://loveholidays.fi) are broken because the CDN domain is getting blocked (`a.loveholidays.com`).
If I get it right, adding the CDN domains to the MDFP list should solve this issue.


dios-david:
Yes sure, now all of them are there.

PR #2602, opened 2020-05-04
Update MDFP list for Discord

tpcstld:
Discord has recently moved its parts of its website from `discordapp.com` to `discord.com`.

We've received reports that our CDN (which remains at `cdn.discordapp.com`), among other things, may be blocked by Privacy Badger on `discord.com`. (https://www.reddit.com/r/discordapp/comments/gdhc83/discord_has_changed_its_domain_to_discordcom/fpherje?utm_source=share&utm_medium=web2x)

Let me know if you need to change something else as well.


tpcstld:
Yes! :D

Embarrassingly, yeah, we do serve some traffic off `discordapp.net` that I didn't know about. Thanks for checking for us!

PR #2627, opened 2020-06-11
Link FAQ on first download page

ablanathtanalba:
This is in an effort to make learning about how Privacy Badger works easier for new users. It features a link to the FAQ page on the first document users see after they download Privacy Badger. Translations for the new messages will take place after this PR is ready to be merged.


ablanathtanalba:
Related to #2624

ablanathtanalba:
The latest commit has the topmost section of the first download page looking like this: 

Screen Shot 2020-06-15 at 8 08 53 AM


And at the end of the tour it has the link to the FAQ as well as an area for contacting us:

Screen Shot 2020-06-15 at 8 09 07 AM

I noticed that the popup preview that's shown in this area is of the old design, so I filed a ticket with E&D for a new image asset that features the most recent redesign of the popup (specifically the buttons, since they're blown up in the image on that first download page) : 
Screen Shot 2020-06-15 at 8 09 23 AM

andresbase:
I like the idea of adding the link to the FAQ.

On the first section of the on-boarding there's some extra text:

> Downloading Privacy Badger for the first time?

I feel like this is redundant and probably could save space by not having that.

<details><summary>Also because now there's the image for the clean browser, the arrow to continue reading is hidden and need to scroll on some mobile devices. It might seem like that's it. Maybe also remove that image?</summary>
<p>

Screen Shot 2020-07-01 at 13 31 30
</p>
</details>

Agreed on updating the look with the new buttons.

<details><summary>I like the small badger and the new text linking to the FAQ. However I feel like there's too much space in between the student badger and the text :</summary>
<p>

Screen Shot 2020-07-01 at 13 05 20
</p>
</details>

Hopefully having a clearer link to the FAQ will resolve some questions and also make it easier for people to reach us.

ablanathtanalba:
Thanks @andresbase ! I removed the two parts you suggested, then reduced the amount of space between the final new image and the text above it.

<details>
  <summary> Here's how it looks now:
Screen Shot 2020-07-05 at 5 42 57 PM
 </summary>
</details>

What do you think?

ghostwords:
Having spent more time thinking about this, I've mostly changed my mind to thinking that it's best to leave the new user welcome page alone and instead update the Help icon in the popup to link to https://privacybadger.org/#faq.

It seems to me at this point that trying to link to the FAQ from the new user welcome page mostly complicates the welcome page without actually solving the original problem of users not being able to get the help they need. For instance, why should users seeking help have to scroll most of the way down through the new user welcome page to find the FAQ link? (Whereas placing the link high up on the page may introduce other problems such as discouraging new users from browsing the welcome page.)

I suggest taking another look through the usability study (see attached to redmine/21097) that prompted this whole effort, in particular its conclusions/recommendations.

ghostwords:
Updating the Help icon link is (A) easy and (B) solves the original problem. However, it removes the ability to reopen the new user welcome page. This may or may not be a problem.

In any case, it seems more important to have users easily find the FAQ (and how to get in touch with us), then for users to reopen the welcome page.

bcyphers:
Agree with @ghostwords last two comments. The new user welcome page doesn't contain any information that isn't already in the FAQ, and users looking for help after they install the extension are more likely to find what they're looking for in the FAQ. Anyone who doesn't see what they're looking for on the welcome page is likely to look to the extension itself next, and the question mark is an obvious next step. 

Let's link the FAQ directly from the extension, and deal with the new user welcome page issue if people complain about it.

ablanathtanalba:
Sounds good to me. Should this branch be scrapped entirely, then, since the only change would be adding the new image? It might be helpful to have links to the FAQ page in both places, no?

Either way, #2642 is open for changing the help button to lead users to the FAQ page.

ghostwords:
>It might be helpful to have links to the FAQ page in both places, no?

There just doesn't seem to be enough benefit for the permanent cost of expanding the welcome page, as well as one-time costs of completing this PR and translating the new strings.

PR #2651, opened 2020-07-21
Improve disabled site popup ui

ablanathtanalba:
In recent conversation it was brought up that the popup UI could be improved on sites that Privacy Badger is disabled on. Specifically that the popup could better reflect the state of PB being "turned off."

There are a couple of proposed changes here. One is greying out just the color of the EFF logo when PB is turned off on a site:

Screen Shot 2020-07-21 at 9 39 56 AM

And the other option greys out both the EFF logo and the little heart on the donate button:

Screen Shot 2020-07-21 at 10 39 26 AM

I'm of no hard opinion either way, but figured I'd present both options here. What's not considered in either of these options is whether or not to also remove the hover effect that changes the border color on the buttons to orange. Is that worth adding as well?

@ghostwords @andresbase

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -438,7 +438,7 @@ function refreshPopup() {
     $("#activate_site_btn").show();
     $("#deactivate_site_btn").hide();
     $("#disabled-site-message").show();
-    $("#title").toggleClass("disabled_site_popup");
+    $("#title").addClass("faded-bw-color-scheme");
   }
 
   // if there is any saved error text, fill the error input with it
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -438,7 +438,8 @@ function refreshPopup() {
     $("#activate_site_btn").show();
     $("#deactivate_site_btn").hide();
     $("#disabled-site-message").show();
-    $("#privacyBadgerHeader").toggleClass("disabled-site-popup");
+    $("#privacyBadgerHeader").toggleClass("disabled_site_popup");
+    $(".ui-icon-heart").attr("style", "color:#808080");
   }
 
   // if there is any saved error text, fill the error input with it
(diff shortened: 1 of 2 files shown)


ghostwords:
Nice! How about a filter of `grayscale(1) opacity(0.4)` just on the "title" div?

ghostwords:
We should do the bare minimum to reinforce the point that Privacy Badger is disabled for the site. No need to remove hover effects or de-emphasize the donation button.

andresbase:
LGTM with @ghostwords' proposals

ablanathtanalba:
I wonder if the contrast ratio on the EFF logo is a concern with a opacity set at .4 : 
Screen Shot 2020-07-22 at 11 56 57 AM

How about set at .6? : 
Screen Shot 2020-07-22 at 11 56 42 AM

ghostwords:
0.6 looks good.

src/js/popup.js
```diff
@@ -438,6 +438,7 @@ function refreshPopup() {
     $("#activate_site_btn").show();
     $("#deactivate_site_btn").hide();
     $("#disabled-site-message").show();
+    $("#title").toggleClass("disabled_site_popup");
```
ghostwords, line 441: Some nits:
- CSS identifiers should be dash-separated (going forward): `disabled-site-popup`
- Maybe calling this class something like "faded" is more to the point?
- Should probably pass `true` as the second argument to `toggleClass()` to be clear that we want to add the class, not remove it. Hmm, we should just use `addClass()` then, right?

ablanathtanalba: good call. did points 1 & 3 in the most recent commit, expanded on point 2 to rename the css class `.faded-bw-color-scheme` to make it more specific

PR #2652, opened 2020-07-23
update popup images on first download page

ablanathtanalba:
Attempting to update the outdated images of the popup on the first download page. My GIMP skills aren't that sharp, so the current images I replaced are a bit blurry compared to the originals. I'll consult with the EFF design folks to figure out how to unblur them.

<details>
<summary>Here's how one appears on the first download page: </summary>
Screen Shot 2020-07-22 at 6 46 33 PM
</details>


ablanathtanalba:
Second commit I included one of the options design gave us. It differs a little slightly from the live one, but at least it isn't blurry like mine are.
<details>
<summary>here's how that looks</summary>
Screen Shot 2020-07-22 at 6 59 28 PM
<details>

ablanathtanalba:
E&D provided some new assets that look much better than the ones I tried duct taping together last week. Here's an updated version of this PR with those assets replacing the old ones. I also rewrote the commit history here to clean it up a bit. I think this looks alright now, but curious to hear what others think -- marking ready for review.

ablanathtanalba:
> I'm not sure about the state of sliders in the first ("learns") image. The current version shows them all as green. The new version shows the "hasn't decided yet" header, which is great, but the domain under the header has been manually blocked?? Instead, show the "hasn't decided yet" header and show some green domains under it (no manual control undo arrows please). We also shouldn't show more red domains above the header than there are green domains under the header. This image is supposed to illustrate learning, not blocking.

I see how that can be confusing, thanks for pointing that out. I'll link E&D to these comments

> So again, match the dimensions of the original?

Ok, I will circle back to E&D and ask them to make the images the same as the dimensions of the previous ones.

> After we get updated images back, could you run them through [imageOptim](https://imageoptim.com/mac) a few times before adding to the repo?

I didn't know about this tool! Cool beans, will launder the next set of assets through it before adding 👍

ablanathtanalba:
Cool! Got some new editions of the images from E+D, I think they look pretty good.

Here's how each looks:

Screen Shot 2020-08-28 at 10 28 01 AM



Screen Shot 2020-08-28 at 10 28 10 AM



Screen Shot 2020-08-28 at 10 28 19 AM





And finally here's how the last asset looks when the screen's width is shrunk down (just before the break point where it restacks the elements to better fit the screen) :

Screen Shot 2020-08-28 at 10 28 37 AM

ablanathtanalba:
> Thank you! Looks good outside of nits.
> 
> I see some editing artifacts that we should be able to remove ourselves:
> 
> * Above `securepubads.g.doubleclick.net` and `c.amazon-adsystem.com` in `catches-trackers.png`
> * To the right of `ad.doubleclick.net` in `disable-badger.png`
> 
> Could you run these through [imageOptim](https://imageoptim.com/mac) a few times until they are closer to current asset file sizes?

Nice catch on those tiny grey marks! I removed them and ran each through ImageOptim as many times as it would allow. Those changes are in the most recent commit.

PR #2659, opened 2020-08-03
Expose privacy overrides in options

### LINKED ISSUE #740: Handle privacy related settings in a more holistic manner

ablanathtanalba:
Fixes #740. 

It exposes an option in the UI for users to control the `alternateErrorPagesEnabled` and `hyperlinkAuditingEnabled` options that PB currently overrides.

### DIFF
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -261,6 +261,14 @@
         "message": "General Settings",
         "description": "This is an options page tab heading."
     },
+    "options_privacy_settings": {
+        "message": "Privacy",
+        "description": "Subheading on the general settings options page."
+    },
+    "options_advanced_settings": {
+        "message": "Advanced",
+        "description": "Subheading on the general settings options page."
+    },
     "intro_next_button": {
         "message": "Take the tour",
         "description": "Intro page welcome button."
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -369,6 +369,10 @@
         "message": "Send websites the \"<a href='https://www.eff.org/issues/do-not-track' target='_blank'>Do Not Track</a>\" signal",
         "description": "Checkbox label for enabling/disabling the Do Not Track signal, found on the general settings page"
     },
+    "options_alternateErrorPagesEnabled_checkbox": {
+        "message": "Re-enable the alternateErrorPagesEnabled Chrome Browser setting",
+        "description": "Checkbox label for enabling/disabling the Do Not Track signal, found on the general settings page"
+    },
     "options_domain_filter_user": {
         "message": "user-controlled",
         "description": "Dropdown control setting on the Tracking Domains options page tab."
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -100,7 +100,7 @@
         "description": "Dropdown control setting on the Tracking Domains options page tab."
     },
     "options_webrtc_setting": {
-        "message": "Prevent WebRTC from leaking local IP address *",
+        "message": "Prevent WebRTC from leaking local IP address",
         "description": "Checkbox label on the general settings page"
     },
     "intro_welcome": {
@@ -148,7 +148,7 @@
         "description": "Checkbox label on the general settings page. Should match wording used in the 'non_tracker' message."
     },
     "options_incognito_warning": {
-        "message": "** Enabling learning in Private/Incognito windows may leave traces of your private browsing history on your computer. By default, Privacy Badger will block trackers it already knows about in Private/Incognito windows, but it won't learn about new trackers. You might want to enable this option if a lot of your browsing happens in Private/Incognito windows.",
+        "message": "Enabling learning in Private/Incognito windows may leave traces of your private browsing history on your computer. By default, Privacy Badger will block trackers it already knows about in Private/Incognito windows, but it won't learn about new trackers. You might want to enable this option if a lot of your browsing happens in Private/Incognito windows.",
         "description": "Detailed explanation shown under checkboxes on the general settings page"
     },
     "report_button": {
@@ -174,7 +174,7 @@
         "description": "Label for a text input box on the Tracking Domains options page tab."
     },
     "options_incognito_setting": {
-        "message": "Learn in Private/Incognito windows **",
+        "message": "Learn in Private/Incognito windows",
         "description": "Checkbox label on the general settings page"
     },
     "show_counter_checkbox": {
@@ -254,7 +254,7 @@
         "description": "Part of a reminder to visit the intro page. Shown in popup until the user clicks on the reminder link or browses through the intro page."
     },
     "options_webrtc_warning": {
-        "message": "* WebRTC can leak your local IP address. Note that enabling this option may degrade performance on web conferencing apps like Google Hangouts.",
+        "message": "WebRTC can leak your local IP address. Note that enabling this option may degrade performance on web conferencing apps like Google Hangouts.",
         "description": "Detailed explanation shown under checkboxes on the general settings page"
     },
     "options_general_settings": {
@@ -370,11 +370,11 @@
         "description": "Checkbox label for enabling/disabling the Do Not Track signal, found on the general settings page"
     },
     "options_alternateErrorPagesEnabled_checkbox": {
-        "message": "Disable Chrome's \"alternateErrorPagesEnabled\" that help resolve navigation errors by sharing your navigation data with a third party service",
+        "message": "Disable sending web addresses to Google to show suggestions when a page can't be found",
         "description": "Checkbox label for enabling/disabling the alternateErrorPagesEnabled option, found on the general settings page"
     },
     "options_hyperlinkAuditingEnabled_checkbox": {
-        "message": "Disable \"<a href='https://www.wilderssecurity.com/threads/major-browsers-to-prevent-disabling-of-click-tracking-privacy-risk.415381/#post-2822723' target='_blank'>hyperlink auditing</a>\" trackers from observing when you click on links",
+        "message": "Disable hyperlink auditing ",
         "description": "Checkbox label for enabling/disabling the hyperlinkAuditingEnabled option, found on the general settings page"
     },
     "options_domain_filter_user": {
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -370,11 +370,11 @@
         "description": "Checkbox label for enabling/disabling the Do Not Track signal, found on the general settings page"
     },
     "options_alternateErrorPagesEnabled_checkbox": {
-        "message": "Disable sending web addresses to Google to show suggestions when a page can't be found",
+        "message": "Disable sending web addresses you visit to Google to show suggestions when a page can't be found",
         "description": "Checkbox label for enabling/disabling the alternateErrorPagesEnabled option, found on the general settings page"
     },
     "options_hyperlinkAuditingEnabled_checkbox": {
-        "message": "Disable hyperlink auditing ",
+        "message": "Disable hyperlink auditing",
         "description": "Checkbox label for enabling/disabling the hyperlinkAuditingEnabled option, found on the general settings page"
     },
     "options_domain_filter_user": {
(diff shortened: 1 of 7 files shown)


src/_locales/en_US/messages.json
```diff
@@ -369,6 +369,10 @@
         "message": "Send websites the \"<a href='https://www.eff.org/issues/do-not-track' target='_blank'>Do Not Track</a>\" signal",
         "description": "Checkbox label for enabling/disabling the Do Not Track signal, found on the general settings page"
     },
+    "options_alternateErrorPagesEnabled_checkbox": {
+        "message": "Re-enable the alternateErrorPagesEnabled Chrome Browser setting",
+        "description": "Checkbox label for enabling/disabling the Do Not Track signal, found on the general settings page"
```
ghostwords, line 374: Description needs to be updated.

src/_locales/en_US/messages.json
```diff
@@ -369,6 +369,10 @@
         "message": "Send websites the \"<a href='https://www.eff.org/issues/do-not-track' target='_blank'>Do Not Track</a>\" signal",
         "description": "Checkbox label for enabling/disabling the Do Not Track signal, found on the general settings page"
     },
+    "options_alternateErrorPagesEnabled_checkbox": {
+        "message": "Re-enable the alternateErrorPagesEnabled Chrome Browser setting",
```
ghostwords, line 373: How does this read to a non-technical Privacy Badger user? "What the heck is alternateErrorPagesEnabled??"

We should look up the language Chrome itself used by finding this setting under advanced privacy settings in an older Chrome version (before Chrome removed this toggle from privacy settings), or perhaps in [Chrome code history](https://source.chromium.org/chromium/chromium/src).

src/js/background.js
```diff
@@ -857,6 +857,13 @@ Badger.prototype = {
     return this.getSettings().getItem("showCounter");
   },
 
+  /**
+  * Check if Re-enable alternateErrorPagesEnabled option is selected
+  */
+  isAlternateErrorPagesEnabled: function() {
```
ghostwords, line 863: We don't need this function.

ablanathtanalba:
> We [set two privacy overrides at this time](https://github.com/EFForg/privacybadger/blob/51c18309848f76f36f8e18c30129ce7d0fed82bd/src/js/background.js#L243-L258). This PR should handle both of them.

Cool beans, agreed. I updated the PR description to reflect this goal, then

src/_locales/en_US/messages.json
ablanathtanalba: I took the wording they use in [extensions privacy docs](https://developer.chrome.com/extensions/privacy) and rephrased it a bit to try and get the point across that a third party is implied. What do you think of: "Disable Chrome's 'alternateErrorPagesEnabled' that help resolve navigation errors by sharing your navigation data with a third party service" ?

src/js/background.js
```diff
@@ -144,6 +146,24 @@ function Badger() {
     return available;
   }
 
+  // checks for availability of alternateErrorPagesEnabled option on browser
+  function checkAlternateErrorPagesSupport() {
```
ghostwords, line 150: I don't think we need these two functions ... We could do these checks where they are needed instead. This would avoid a bunch of unnecessary code. Instead of defining, looking up, message passing ... do the check directly on the options page.

src/js/background.js
```diff
@@ -144,6 +146,24 @@ function Badger() {
...
+    if (!chrome.privacy.services.alternateErrorPagesEnabled) {
+      return false;
+    } else {
+      return true;
+    }
+  }
+
+  // checks for availability of hyperlinkAuditingEnabled option on browser
+  function checkHyperlinkAuditingSupport() {
```
ghostwords, line 159: Same as above.

src/js/background.js
```diff
@@ -843,6 +865,14 @@ Badger.prototype = {
     return this.getSettings().getItem("checkForDNTPolicy");
   },
 
+  isAlternateErrorPagesEnabled: function() {
```
ghostwords, line 868: These are unused. Also unnecessary as a rule: look up from `getSettings()` directly where you need it unless we do the lookup enough to justify its own function.

src/js/options.js
```diff
@@ -114,6 +114,34 @@ function loadOptions() {
   $("#enable_dnt_checkbox").prop("checked", OPTIONS_DATA.isDNTSignalEnabled);
   $("#check_dnt_policy_checkbox").on("click", updateCheckingDNTPolicy);
   $("#check_dnt_policy_checkbox").prop("checked", OPTIONS_DATA.isCheckingDNTPolicyEnabled).prop("disabled", !OPTIONS_DATA.isDNTSignalEnabled);
+  $("#alternateErrorPagesEnabled_checkbox").on("click", toggleAlternateErrorPagesSetting);
+  $("#alternateErrorPagesEnabled_checkbox").prop("checked", OPTIONS_DATA.isAlternateErrorPagesEnabled);
+  $("#hyperlinkAuditingEnabled_checkbox").on("click", toggleHyperlinkAuditingSetting);
+  $("#hyperlinkAuditingEnabled_checkbox").prop("checked", OPTIONS_DATA.isHyperlinkAuditingEnabled);
```
ghostwords, line 120: `OPTIONS_DATA.isHyperlinkAuditingEnabled` is never set.

src/js/options.js
```diff
@@ -114,6 +114,34 @@ function loadOptions() {
   $("#enable_dnt_checkbox").prop("checked", OPTIONS_DATA.isDNTSignalEnabled);
   $("#check_dnt_policy_checkbox").on("click", updateCheckingDNTPolicy);
   $("#check_dnt_policy_checkbox").prop("checked", OPTIONS_DATA.isCheckingDNTPolicyEnabled).prop("disabled", !OPTIONS_DATA.isDNTSignalEnabled);
+  $("#alternateErrorPagesEnabled_checkbox").on("click", toggleAlternateErrorPagesSetting);
+  $("#alternateErrorPagesEnabled_checkbox").prop("checked", OPTIONS_DATA.isAlternateErrorPagesEnabled);
```
ghostwords, line 118: `OPTIONS_DATA.isAlternateErrorPagesEnabled` is never set.

src/js/options.js
```diff
@@ -114,6 +114,34 @@ function loadOptions() {
...
+  $("#hyperlinkAuditingEnabled_checkbox").on("click", toggleHyperlinkAuditingSetting);
+  $("#hyperlinkAuditingEnabled_checkbox").prop("checked", OPTIONS_DATA.isHyperlinkAuditingEnabled);
+
+  // only show the alternateErrorPagesEnabled checkbox if browser supports it
+  if (!OPTIONS_DATA.alternateErrorPagesAvailable) {
```
ghostwords, line 123: Instead hide by default and show if `chrome.privacy` and `chrome.privacy.services` and `chrome.privacy.alternateErrorPagesEnabled` all exist?

src/js/options.js
```diff
@@ -114,6 +114,34 @@ function loadOptions() {
...
+    // check the select box if it is already disabled in the browser
+    chrome.privacy.services.alternateErrorPagesEnabled.get({}, result => {
+      if (result.value == false) {
+        $('#alternateErrorPagesEnabled_checkbox').prop("checked", true);
+      }
+    });
+  }
+
+  // only show the hyperlinkAuditingEnabled checkbox if browser supports it
+  if (!OPTIONS_DATA.hyperlinkAuditingAvailable) {
```
ghostwords, line 135: Same as above.

src/js/options.js
```diff
@@ -737,6 +765,40 @@ function toggleWebRTCIPProtection() {
...
+// handles toggling the alternateErrorPagesEnabled setting
+function toggleAlternateErrorPagesSetting() {
+  // ensure this is only attempting to be set on supportive browsers
+  if (!OPTIONS_DATA.alternateErrorPagesAvailable) {
+    return;
+  }
+
+  let cps = chrome.privacy.services;
+
+  // whatever the current setting is at, reverse it
```
ghostwords, line 777: We should instead `clear({})` the setting. It doesn't make sense for us to continue to override the setting (to be enabled) if the user disables our override for whatever reason. We should relinquish control instead.

src/js/options.js
```diff
@@ -737,6 +765,40 @@ function toggleWebRTCIPProtection() {
...
+// handles toggling the hyperlinkAuditingEnabled setting
+function toggleHyperlinkAuditingSetting() {
+  // ensure this is only attempting to be set on supportive browsers
+  if (!OPTIONS_DATA.hyperlinkAuditingAvailable) {
+    return;
+  }
+
+  let cpw = chrome.privacy.websites;
+
+  //whatever the current setting is at, reverse it
```
ghostwords, line 794: Same as above.

src/js/webrequest.js
```diff
@@ -919,8 +919,10 @@ function dispatcher(request, sender, sendResponse) {
     }
 
     sendResponse({
+      alternateErrorPagesAvailable: badger.alternateErrorPagesAvailable,
```
ghostwords, line 922: Don't need these two.

ablanathtanalba:
Coming back to this, I wonder if it's a better strategy to remove the privacy overrides option that's currently in the background page script, and just keep it in the options page one. They're the only two privacy overrides that are set in `background.js`, so it would remove a good chunk of code and keep the options page logic all in one place, rather than creating a special exception for these two. What do you think, @ghostwords ?

ghostwords:
We want to keep applying privacy overrides (after taking user preferences into account) on Privacy Badger startup so that it is more likely privacy overrides are actually set.
- Privacy Badger might have been disabled, which would have unset our browser privacy overrides. Re-enabling Privacy Badger should restore the overrides.
- Perhaps an override could not be applied in the past because another extension used to control the browser setting

ablanathtanalba:
Ok, the newest commit makes a small tweak in the initialization phase that checks the user preferences for how to handle these options and continue on with privacy overrides during badger initialization.

On another note, I ran into some confusing workflow when it comes to disabling this option, aka re-enabling these privacy options on the browser. I think it's misleading to just relinquish control of these settings when the user unselects the checkboxes in the options page; relinquishing control by running `.clear({})` instead of `.set({...})` isn't clearly described in UI options page, and I don't think there's a way to do so without clogging the space with a bunch more description text. For instance, when control is relinquished and the browser preference is cleared, it doesn't necessarily mean that the value is set to `true`. I think it would take disabling the checkbox or more info tooltip text to try and describe this situation... which I think we are trying to move away from overcomplicating, right? If the user decides to override these settings elsewhere, in the browser or with some other browser extension, the current working state of this feature will update the options page checkbox to show how the preference is currently set (and toggling it would update the settings storage to its current state if in that case there is some discrepancy).

src/js/options.js
ablanathtanalba: I feel like going this route could be misleading to the user.

I'm not sure there's a way to describe this to the user in the options page. For instance, when control is relinquished and the browser preference is cleared, it doesn't necessarily mean that the value is set to `true`. I think it would take disabling the checkbox or adding more info tooltip text to try and describe this situation... which I think we are trying to move away from overcomplicating, right? If the user decides to override these settings elsewhere, in the browser or with some other browser extension, the current working state of this feature will update the options page checkbox to show how the preference is currently set (and toggling it would update the settings storage to its current state if in that case there is some discrepancy).

ghostwords: Our privacy overrides take the form of "Disable INVASIVE_SETTING". If the user disables an override, they tell Privacy Badger to stop disabling INVASIVE_SETTING. I think it's clear the checkbox doesn't ever enable INVASIVE_SETTING.

ghostwords:
Here is the latest screenshot of my WIP:

Screenshot from 2020-08-31 17-53-37

ablanathtanalba:
That looks really good! The red warning triangles have the same popup tooltip when hovered on?

ghostwords:
Thanks! Yes, and probably should be on click on mobile.

ghostwords:
OK, let's merge and later continue with #1897 (add an override for `chrome.privacy.network.networkPredictionEnabled`).

PR #2693, opened 2020-10-01
Implement global privacy control

bcyphers:
This is a quick, simple implementation of sec-gpc, the upcoming do-not-sell-or-share standard for CCPA/nevada privacy/GDPR purposes. The signal looks an awful lot like DNT. 

With this PR, Sec-GPC will be sent whenever DNT is: the global checkbox in the options page that enabled DNT now enables both signals, and Privacy Badger will not send Sec-GPC to sites on the disabled list.

Like DNT, there are two ways the signal is conveyed: through a browser header ("Sec-GPC: 1"), and through a navigator property (`navigator.globalPrivacyControl = 1`). This PR sets GPC to 1 by default in both of those channels. Both are controlled by the `badger.sendDNTSignal` setting.

In the future, it might make sense to separate out DNT and GPC, but this is the simplest way I could think to implement. Most of the work that still needs to be done is around copy: we should let users know about GPC in the FAQ and the start page.

### DIFF
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -374,8 +374,8 @@
         "description": ""
     },
     "options_enable_dnt_checkbox": {
-        "message": "Send websites the \"<a href='https://www.eff.org/issues/do-not-track' target='_blank'>Do Not Track</a>\" signal",
-        "description": "Checkbox label for enabling/disabling the Do Not Track signal, found on the general settings page"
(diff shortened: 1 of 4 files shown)


bcyphers:
We should probably add unit tests for this as well. For now, it can be tested here: http://global-privacy-control.glitch.me/

ghostwords (approved):
Thank you!

PR #2732, opened 2021-01-13
add a dark theme

dannycolin:
This adds a dark theme for the popup, settings page and
first run page.

Fixes: #2630, closes #2648

A few elements, like on the light theme, aren't themed (e.g checkboxes) because they need `appearance: none` and to redo the widget. So I think we could if needed do this in a different patch. Any thought?


ghostwords (approved):
Thank you, looks great!

PR #2787, opened 2021-07-22
Make disable on site more prominent

ablanathtanalba:
Related to #2781, #2021. Supersedes #2660.

This makes some more cosmetic changes to the popup to make the disable and enable buttons more prominent.

* adds a colored "power" icon to make it more visually clear what the button does
* simplifies the language on the button by removing "Privacy Badger" from the message (it's obvious that the control is for Privacy Badger)
* changes the disable and enable buttons to be stacked at 100% width with the other two buttons below sharing a single row.

Screen Shot 2021-07-21 at 5 28 20 PM

### DIFF
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -218,8 +218,8 @@
         "description": "Tooltip that comes up when you hover over the 'tracking domains' link under the Tracking Domains tab on the options page."
     },
     "intro_report_button": {
-        "message": "Please don't forget to click on 'Did Privacy Badger break this site'. We respect your privacy so we don't send automatic reports.",
-        "description": "Intro page paragraph. The quoted message ('Did Privacy Badger break this site') should match the first part of the translation for the 'report_broken_site' message."
+        "message": "Please don't forget to click on 'Report broken site'. We respect your privacy so we don't send automatic reports.",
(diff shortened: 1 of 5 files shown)


ablanathtanalba:
Remaining question:
Should the "power" icon switch to grey when in powered-off state? Or remain orange as is?

Screen Shot 2021-07-22 at 11 04 27 AM

ablanathtanalba:
Thanks for these tweaks!

I'm definitely in favor of simplifying the report button message as well. That's a keeper.

I'm on the fence about the rounded corners for the buttons. On the one hand, it does kinda fall in line with how many other web things style their buttons, so it appeals to that sense of familiarity with people. On the other hand, the curve radius on the button is just different enough from the highlighted blocked resources outlines that it's noticeable and kind of bothersome. 
<details>
<summary> example screenshot</summary>
Screen Shot 2021-07-23 at 11 18 50 AM
</details>

I'm not in favor of the drop shadow on the power button, however. I think it's a little visually confusing when you see it, since it's the only thing in the popup that would have that kind of lifted dimension.

ablanathtanalba:
What if the button class styling didn't change much, but we emphasized the power button icon a bit more?

Screen Shot 2021-07-23 at 11 30 13 AM

ghostwords:
Latest revision:

Screenshot from 2021-07-26 15-15-39

PR #2816, opened 2021-11-08
Update yellowlist.txt for additional Google map tile servers

camwebb:
Google's Map Tile servers `mt0.google.com` and `mt1.google.com` are in the yellow list, but sometimes `mt2.google.com` and `mt3.google.com` are also used to server map tiles, and should therefore also be in the yellow list. Example of a failing site: <https://map.treetracker.org/>. Thanks.


ghostwords (approved):
Hello and thank you!

PR #2835, opened 2022-04-04
Create a special error message for Firefox users

ablanathtanalba:
Related to #2750 and #2804

This adds a line into the block of error text that users get in the popup, and creates another link to direct users to the FAQ entry on the website where we describe how to fix the common memory error they see and otherwise get an unhelpful "UNEXPECTED ERROR OCCURRED" message.

Hopefully we can remove this and the other related minor fixes in the future when this is no longer an issue for Firefox users with low disk space.

### DIFF
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -265,6 +265,20 @@
         "message": "Please <a href='https://www.eff.org/privacybadger#faq-I-found-a-bug!-What-do-I-do-now?' target='_blank'>tell us</a> about the following error:",
         "description": "Shown in the popup when there is a problem with the user's Privacy Badger extension that we want to encourage the user to tell us about."
     },
+    "popup_error_text_low_disk_space_issue": {
+      "message": "Failed to write to extension storage: '$ERROR_TEXT$'. Is your disk low on space? $LEARN_MORE_LINK$",
+      "description": "Special error message for the 'Unkown Error Occurred' text that sometimes occurs on Firefox browsers with low disk space",
+      "placeholders": {
+        "error_text": {
+          "content": "$1",
+          "example": "Unkown error occurred"
+        },
+        "learn_more_link": {
+          "content": "$2",
+          "example": "Visit <a href='https://eff.org/'>here</a> to learn more."
+        }
+      }
+    },
     "firstRun_title": {
         "message": "Thank you for installing Privacy Badger!",
         "description": "Title and header of the intro page."
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -266,16 +266,12 @@
         "description": "Shown in the popup when there is a problem with the user's Privacy Badger extension that we want to encourage the user to tell us about."
     },
     "popup_error_text_low_disk_space_issue": {
-      "message": "Failed to write to extension storage: '$ERROR_TEXT$'. Is your disk low on space? $LEARN_MORE_LINK$",
-      "description": "Special error message for the 'Unkown Error Occurred' text that sometimes occurs on Firefox browsers with low disk space",
+      "message": "Failed to write to extension storage: '$ERROR_TEXT$'. Is your disk low on space? Visit <a href='https://www.privacybadger.org/#faq-I-found-a-bug!-What-do-I-do-now?'>privacybadger.org</a> to learn more.",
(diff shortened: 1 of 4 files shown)


src/_locales/en_US/messages.json
```diff
@@ -265,6 +265,20 @@
         "message": "Please <a href='https://www.eff.org/privacybadger#faq-I-found-a-bug!-What-do-I-do-now?' target='_blank'>tell us</a> about the following error:",
         "description": "Shown in the popup when there is a problem with the user's Privacy Badger extension that we want to encourage the user to tell us about."
     },
+    "popup_error_text_low_disk_space_issue": {
+      "message": "Failed to write to extension storage: '$ERROR_TEXT$'. Is your disk low on space? $LEARN_MORE_LINK$",
+      "description": "Special error message for the 'Unkown Error Occurred' text that sometimes occurs on Firefox browsers with low disk space",
```
ghostwords, line 270: "An unexpected error occurred", not "unknown"

src/_locales/en_US/messages.json
```diff
@@ -265,6 +265,20 @@
...
+      "placeholders": {
+        "error_text": {
+          "content": "$1",
+          "example": "Unkown error occurred"
```
ghostwords, line 274: "An unexpected error occurred"

src/js/popup.js
```diff
@@ -84,6 +84,20 @@ function showNagMaybe() {
 
   function _showError(error_text) {
     $('#instruction-text').hide();
+    $('#error-message').text(error_text);
+    // if error is the 'unexpected error occurred' Firefox issue commonly associated with low disk space, show special message
+    if (typeof browser == "object" && typeof browser.runtime.getBrowserInfo == "function" && error_text.includes("unexpected")) {
```
ghostwords, line 89: Let's look for the exact entire string ("An unexpected error occurred" I think) and remove the browser detection. It's unlikely a different browser will have this exact same error string, but if it does happen, there is no big problem and we will eventually learn of it and reassess.

The browser detection here complicates the logic for not enough benefit in my estimate. The browser lookup is async; there is a brief flash of the "let us know" UI before it gets hidden.

src/js/popup.js
```diff
@@ -84,6 +84,20 @@ function showNagMaybe() {
...
+      browser.runtime.getBrowserInfo().then(function (info) {
+        if (info.name == "Firefox") {
+          // fetch the "learn more" message and populate with our FAQ link
+          let learn_more_link = chrome.i18n.getMessage("learn_more_link", ["<a href='https://www.eff.org/privacybadger#faq-I-found-a-bug!-What-do-I-do-now?'>our FAQ</a>"]);
```
ghostwords, line 93: Let's use the privacybadger.org link, not the old style eff.org/privacybadger link.

ghostwords, line 93: "our FAQ" is hardcoded in English. Let's bypass adding another message by using "privacybadger.org" here.

src/js/popup.js
```diff
@@ -84,6 +84,20 @@ function showNagMaybe() {
...
+          // replace current error text with this special message
+          $('#error-text').html(chrome.i18n.getMessage("popup_error_text_low_disk_space_issue", [error_text, learn_more_link]));
```
ghostwords, line 95: Style `error_text` "warning" red?

src/js/popup.js
ghostwords, line 93: The link is unstyled. Let's also make sure to check styles in dark mode.

ghostwords (commented):
Thank you! Please see my feedback inline.

ablanathtanalba:
Alright, finally got back to follow up with some cleanup on this branch, hitting the feedback you left here @ghostwords. Thanks for that. I'm still having some difficulty replicating the low disk space environment to trigger the error message (though at this point it might be safe to just falsely trigger it as always true if this is just for styling purposes at this point...?) -- may need to sidechannel with you for your recommendations on how you went about doing that last time.

ablanathtanalba:
Screen Shot 2022-08-01 at 9 42 04 AM


Simplified and cleaned up a bunch. Here's what it looks like. Welcoming all thoughts! Here is what a regular error message looks like, in comparison (i personally think the button is a bit much, but it's a better call to action than the FAQ link in the above example ought to be)

Screen Shot 2022-08-01 at 7 35 15 AM

PR #2983, opened 2024-06-13
Show Chromium users how to pin extension in toolbar

lenacohen:
Encourage Chrome users to pin Privacy Badger to their toolbar by:
1) Adding information about how and why to pin the extension
2) Adding an arrow pointing to where they can click to pin the extension

Test by running branch in Chrome to see changes `(make runch)` and another browser to confirm nothing changed `(make runff)`

Seeking feedback on the design and language, especially for the arrow. 

Also, following up on #2959 and d1a748cec47ebad30e3d49c8356fcdb04b35ecd3, I fixed a UI bug I noticed with the Share icons on narrow screens:
Screenshot 2024-06-13 at 6 13 03 PM

Part of #2781


ghostwords (commented):
Thank you! First impressions:

- The nudge needs to support dark mode. The text is invisible at the moment.
- The nudge would ideally account for scrollbars. My scrollbar doesn't autohide so the tail/pointer is off by the scrollbar's amount (I think). Not a blocking issue, but would be nice if we could handle this properly. Dunno yet if there is a good way for us to do so.
- We should focus on teaching the user how to disable Privacy Badger. Something like, 'Pin ~~this extension~~Privacy Badger to your toolbar for easy access ~~. The Privacy Badger icon will show you how many trackers it blocks on each page. Click on the icon to manage your privacy settings.~~ to the "Disable for this site" button.' But a bit more than this -- let's think about how to combine the "If you think Privacy Badger is breaking a page" text at the bottom with this nudge. The big idea is that we used to tell users about disabling Privacy Badger way down on the welcome page and now we want to do it at the top. Let's also review our previous work in #2812 in light of all of this.

src/skin/js/firstRun.js
```diff
@@ -15,4 +15,18 @@ $(function () {
       });
     }
   });
+
+  // Nudge Chrome users to pin the extension to their toolbar
+  if (navigator && navigator.userAgent && navigator.userAgent.match(/chrome/i)) {
```
ghostwords, line 20: Our goal here is to not show the nudge in browsers where we can [auto-pin using the `default_area` property](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action#syntax), right? That's only Firefox. All Chromium-based browsers (Edge, Opera, Brave, ...) should also get this prompt.

We should avoid checking the UA whenever possible because users change their UA or blank it out among other reasons.

`browser.browserAction.getUserSettings()` exists in Firefox and will async return an object with `isOnToolbar` set to `true`.
`chrome.action.getUserSettings()` exists in MV3 Chrome and will async return an object with `isOnToolbar` set to `false`.

We should probably use this to decide whether to show the nudge or not.

lenacohen: I thought this change was only for Chrome users but this is a good idea for how to target all browsers except for Firefox! Thank you! 

One complication with including Edge, Opera, and Brave is that the icons that users need to click are a little different in each browser. For example, Opera has a box instead of a puzzle piece as the extension icon in the toolbar. To pin on Edge, users select an eye icon instead of a pin icon. So we might need to detect which browser the page is on and load a different image for each browser. That's what Daly did in [this PR](https://github.com/EFForg/privacybadger/pull/2812/files#diff-4488bb389c2092a3e38367777d8fc45e9a422f51943ff728a8901cb490716a3a). 

Let's discuss what approach to take for different browsers when we talk about a new design proposal?

lenacohen:
I don't love the popup message but I'm still thinking about alternatives. I'm worried the "Disable for this site" reference needs more context or is too specific, since it's not all the popup has. Here's one alternative that's a bit longer: 
- Pin Privacy Badger to your toolbar for easy access to your tracker-blocking settings and the "Disable for this site" button.

src/skin/firstRun.html
```diff
@@ -28,6 +28,17 @@ <h1 class="i18n_name"></h1>
...
 
+    <!-- arrow showing users where to pin their extension, only shown on chromium -->
+    <div id="pin-nudge-arrow">
+      <div>
+        <a href="" id="dismiss-nudge" role="button" aria-label="i18n_report_close">
+          <img src="../icons/close.svg" alt="">
+        </a>
+        <p id="pin-nudge-text"><span id="pin-nudge-cta" class="i18n_intro_pin_action"></span> <span class="i18n_intro_pin_purpose"></span></p>
+      </div>
+      <img id="pin-image" src="images/pinning-instructions-4.png" alt="TO DO">
```
ghostwords, line 39: The alt text should be blank if this is a decorative image, and should properly describe what it is we mean to convey otherwise. So here we should probably say something like:

To pin Privacy Badger's icon, click on the "puzzle piece" icon in your browser toolbar. Then, click on the "pin" next to Privacy Badger.

src/skin/firstRun.html
```diff
@@ -28,6 +28,17 @@ <h1 class="i18n_name"></h1>
       </div>
     </header>
 
+    <!-- arrow showing users where to pin their extension, only shown on chromium -->
+    <div id="pin-nudge-arrow">
```
ghostwords, line 32: Let's do the `display:none` with an inline style tag here in the HTML.

src/_locales/en_US/messages.json
```diff
@@ -359,6 +359,22 @@
         "message": "Privacy Policy",
         "description": "Shown at the bottom of the intro page."
     },
+    "intro_easy_to_manage": {
```
ghostwords, line 362: Unused?

src/_locales/en_US/messages.json
```diff
@@ -359,6 +359,22 @@
...
+        "message": "Easy to manage",
+        "description": "Intro page paragraph heading on Chrome."
+    },
+    "intro_easy_to_manage_paragraph": {
```
ghostwords, line 366: Unused?

src/_locales/en_US/messages.json
```diff
@@ -359,6 +359,22 @@
...
+        "message": "Pin this extension to your toolbar for easy access. The Privacy Badger icon will show you how many trackers it blocks on each page. Click on the icon to manage your privacy settings.",
+        "description": "Intro page paragraph on Chrome."
+    },
+    "intro_pin_action": {
+        "message": "Pin Privacy Badger to your toolbar",
+        "description": "Shown on arrow pointing at Chrome extension icon"
```
ghostwords, line 372: The translator probably wants to know how to find this. Also, no longer an arrow. "Part of a Chrome-only intro page overlay"?

src/_locales/en_US/messages.json
```diff
@@ -359,6 +359,22 @@
...
+    },
+    "intro_pin_purpose": {
+        "message": "for easy access to the \"Disable for this site\" button.",
```
ghostwords, line 375: We shouldn't split up locale messages like this, as it makes it harder for translators to make sense of what they are translating, and there may be grammar and other locale difference that don't work well with the way we've chosen to split the message.

Let's use a single message with placeholders instead.

src/skin/css/firstRun.css
```diff
@@ -498,7 +498,6 @@ header .title-bar h1 {
   display:block;
   margin:2.5rem auto 2rem;
   padding-bottom:3rem;
-  width:94px;
```
ghostwords, line 501: Why did we remove this?

src/skin/firstRun.html
ghostwords: Don't need the "-4" in the name?

src/skin/js/firstRun.js
```diff
@@ -15,4 +15,33 @@ $(function () {
       });
     }
   });
+
+  // Don't show the pin nudge in Firefox because extensions are pinned automatically
+  // chrome.action is only available in MV3 and chrome.browserAction is only available in <= MV2
+  // chrome.browserAction.getUserSettings doesn't exist in MV2 but does in Firefox
+  var isChromiumMv3 = chrome.action;
```
ghostwords, line 22: Let's use `let` over `var` going forward. Maybe `const` sometimes, but mainly `let`.

src/skin/js/firstRun.js
```diff
@@ -15,4 +15,33 @@ $(function () {
...
+  if (isChromiumMv2 || isChromiumMv3) {
+    $("#pin-nudge-arrow").show();
+  }
+
+  $("#dismiss-nudge").on("click", function (e) {
+    e.preventDefault();
+    $("#pin-nudge-arrow").hide();
+  });
+
+  let intervalId;
```
ghostwords, line 33: If it's not an object, just an int, let's call it `interval_id` then.

src/skin/css/firstRun.css
```diff
@@ -789,6 +795,53 @@ body {
...
+
+#pin-nudge-cta {
+  font-family:"OpenSans-Bold",sans-serif;
+}
+
+#pin-image {
+  width: 250px;
+}
+
+#pin-nudge-arrow {
```
ghostwords, line 828: Remove "-arrow" from the name?

ghostwords (commented):
Thank you! Left some feedback inline, plus:

- Is the puzzle piece icon also unused?
- The overlay show appear on the left on RTL locales. Let's flip the image in the overlay too.
- Does running any new images through https://imageoptim.com/mac help save space?
- We could also use a placeholder for "Disable for this site" in  '... access to the "Disable for this site" button'. Then, we won't have to worry about having the translation here and in the pop being in sync.
- I wonder if we should gray out the rest of the page to draw attention to, and to force interaction with the overlay. Kind of like what happens in the popup when `badger.criticalError` is set in the background process. I am not sure if we should, just thinking whether the overlay is too easy to ignore. The orange hero background, plus the "Take the tour" CTA button, plus the social sharing buttons vs. this gray overlay, you know?

src/skin/firstRun.html
```diff
@@ -28,6 +28,17 @@ <h1 class="i18n_name"></h1>
...
+      <div>
+        <a href="" id="dismiss-nudge" role="button" aria-label="i18n_report_close">
+          <img src="../icons/close.svg" alt="">
```
ghostwords, line 35: Why do we need a new close.svg when we already have a similar "X" to close elsewhere?

src/skin/css/firstRun.css
lenacohen: I removed to solve a UI bug with share links on narrow windows (see the PR description for a screenshot). Removing the 94px width fixed this UI bug and didn't introduce new bugs on all the screen sizes I checked, so I'm not sure why it was there

src/skin/firstRun.html
lenacohen: Oops yes thanks for the alt text, I forgot to come back to my TODO!

src/skin/firstRun.html
lenacohen: close.svg is basically the same as the inline svg! I meant to replace the inline svg with the re-usable icon, since it'll reduce lines of code and increase readability. is that okay with you?

src/_locales/en_US/messages.json
lenacohen: I split the locale messages so that I could apply different font-families to each. I'll try doing the same thing with placeholders!

lenacohen:
> Thank you! Left some feedback inline, plus:
> 
> * Is the puzzle piece icon also unused?
> * The overlay should appear on the left w/ RTL locales. Let's flip the image in the overlay too.
> * Does running any new images through https://imageoptim.com/mac help save space?
> * We could also use a placeholder for "Disable for this site" in  '... access to the "Disable for this site" button'. Then, we won't have to worry about having the translation here and in the popup being in sync.
> * I wonder if we should gray out the rest of the page to draw attention to, and to force interaction with the overlay. Kind of like what happens in the popup when `badger.criticalError` is set in the background process. I am not sure if we should, just concerned whether the overlay is too easy to ignore. The orange hero background, plus the "Take the tour" CTA button, plus the social sharing buttons vs. the gray overlay, you know?
> * Maybe the X should be smaller?

Thanks for the feedback! I removed the unused puzzle piece icon and ran the new images through imageoptim. I like the gray background while the nudge is showing!

src/skin/js/firstRun.js
```diff
@@ -15,4 +15,45 @@ $(function () {
...
+
+  function hideNudgeOverlay() {
+    $("#pin-nudge").fadeOut();
+    $("#overlay").fadeOut();
+  }
+
+  // Don't show the pin nudge in Firefox because extensions are pinned automatically
+  // chrome.action is only available in MV3 and chrome.browserAction is only available in <= MV2
+  // chrome.browserAction.getUserSettings doesn't exist in MV2 but does in Firefox
+  let isChromiumMv3 = chrome.action;
```
ghostwords, line 27: `!!chrome.action` to make a boolean

ghostwords: [Booleans should get snake case names](https://github.com/EFForg/privacybadger/blob/master/doc/develop.md#style-guide), same as integers.

src/_locales/en_US/messages.json
```diff
@@ -359,6 +359,22 @@
         "message": "Privacy Policy",
         "description": "Shown at the bottom of the intro page."
     },
+    "intro_pin_nudge": {
+        "message": "$START_SPAN_TAG$First, pin Privacy Badger to your toolbar$END_SPAN_TAG$ for easy access to the \"$DISABLE_FOR_SITE$\" button.",
+        "description": "Part of a Chrome-only overlay on the intro page",
+        "placeholders": {
+            "start_span_tag": {
+                "content": "<span style='font-family:\"OpenSans-Bold\",sans-serif;'>"
```
ghostwords, line 367: Let's use the `pin-cta` ID or some other CSS identifier instead. Shorter and less likely for us to have to edit locales when making changes in the future.

src/_locales/en_US/messages.json
```diff
@@ -359,6 +359,22 @@
         "message": "Privacy Policy",
         "description": "Shown at the bottom of the intro page."
     },
+    "intro_pin_nudge": {
+        "message": "$START_SPAN_TAG$First, pin Privacy Badger to your toolbar$END_SPAN_TAG$ for easy access to the \"$DISABLE_FOR_SITE$\" button.",
```
ghostwords, line 363: Should "First," be bolded? Not sure, just wondering which one looks/reads better.

src/skin/firstRun.html
```diff
@@ -28,6 +28,18 @@ <h1 class="i18n_name"></h1>
       </div>
     </header>
 
+    <!-- Overlay nudging users to pin their extension (only shown on chromium browsers) -->
+    <div id="overlay" style="display:none;" class="nudge-overlay"></div>
+    <div id="pin-nudge" style="display:none;" class="nudge-overlay">
```
ghostwords, line 33: Why is the nudge-overlay class for?

src/skin/firstRun.html
```diff
@@ -28,6 +28,18 @@ <h1 class="i18n_name"></h1>
...
+      <div>
+        <a href="" id="dismiss-nudge" role="button" aria-label="i18n_report_close">
+          <img src="../icons/close.svg" alt="">
+        </a>
+        <p id="pin-nudge-text"></p>
+      </div>
+      <img id="pin-image" src="images/pinning-instructions.png" alt="To pin Privacy Badger's icon, click on the &quot;puzzle piece&quot; icon in your browser toolbar. Then, click on the &quot;pin&quot; icon next to Privacy Badger.">
```
ghostwords, line 40: Let's i18n the alt text.

src/skin/js/firstRun.js
```diff
@@ -15,4 +15,45 @@ $(function () {
...
+  async function checkIsPinned() {
+    let userSettings = await chrome.action.getUserSettings();
+    if (userSettings.isOnToolbar) {
+      hideNudgeOverlay();
+      clearInterval(interval_id);
+    }
+  }
+
+  // For Chromium, we can only check if PB is pinned in MV3
+  if (chrome.action) {
```
ghostwords, line 56: Should reuse the boolean we already made?

ghostwords (commented):
Some more nits and a question.

src/_locales/en_US/messages.json
lenacohen: I originally thought it looked better with "First" bolded but now I think the opposite! What do you think? I don't have a strong preference

Screenshot 2024-07-01 at 11 13 34 AM
Screenshot 2024-07-01 at 11 14 04 AM

src/skin/firstRun.html
lenacohen: just something I forgot to remove! thanks for catching

src/_locales/en_US/messages.json
lenacohen: good point!

src/_locales/en_US/messages.json
ghostwords: Same ...

ghostwords (approved):

PR #3034, opened 2024-11-21
Add close icon to breakage note tooltip

lenacohen:
As we are making the breakage tooltip appear consistently (#3029), we should follow [tooltip best practices](https://userpilot.com/blog/tooltip-best-practices/#:~:text=Allow%20easy%20exit%20from%20tooltips&text=You%20could%20use%20an%20X,user%20knows%20how%20to%20exit.) and make it clear how users can close it

Screenshot 2024-11-21 at 10 51 29 AM

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -569,7 +569,12 @@ function createBreakageNote(domain, i18n_message_key) {
   // return the tooltip we want (the breakage note, not Allow)
   $slider_allow.tooltipster('destroy').tooltipster({
     autoClose: false,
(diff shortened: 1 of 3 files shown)


ghostwords (commented):
What does this look like when using the most verbose locale?

lenacohen:
> What does this look like when using the most verbose locale?

The text goes from 3 lines to 4 lines. I think it looks okay but also might be able to prevent that by making the icon smaller and closer to the corner.

Снимок экрана 2024-11-21 в 12 53 38 PM

Снимок экрана 2024-11-21 в 12 52 09 PM

PR #3065, opened 2025-02-27
Fix widget styling bugs

lenacohen:
- Fixes https://github.com/EFForg/privacybadger/issues/3024
- Move widget placeholder inside shadow DOM to avoid interferences
  - Fixes placeholder being hidden on https://developer.chrome.com/blog/extension-news-january-2025
  - Fixes widget overlapping with navbar on page scroll on https://www.msn.com/en-us/sports/nfl/super-bowl-mvp-jalen-hurts-had-lesbian-nation-behind-him-for-the-biggest-night-of-his-life/ar-AA1yLY0D
- Address widget replacement overlapping with content by ensuring that the parent node height is at least as large as the min height of the widget
  - Fixes overlap on https://www.wausaudailyherald.com/story/news/politics/elections/2024/09/27/heres-what-we-know-so-far-about-wausau-mayor-doug-dinys-removal-of-the-clerks-drop-box/75402110007/
- Remove code that the shadow DOM makes unnecessary 
  - Computed styling is still necessary on some pages: https://www.wausaudailyherald.com/story/news/politics/elections/2024/09/27/heres-what-we-know-so-far-about-wausau-mayor-doug-dinys-removal-of-the-clerks-drop-box/75402110007/

### DIFF
--- a/src/js/contentscripts/socialwidgets.js
+++ b/src/js/contentscripts/socialwidgets.js
@@ -408,6 +408,9 @@ function createReplacementWidget(widget, elToReplace) {
   let name = widget.name;
 
   let widgetFrame = document.createElement('iframe');
+  let shadowHost = document.createElement("div");
+  let shadowRoot = shadowHost.attachShadow({ mode: "closed" });
+  shadowRoot.appendChild(widgetFrame);
 
   // widget replacement frame styles
   let border_width = 1;
@@ -648,7 +651,7 @@ function createReplacementWidget(widget, elToReplace) {
   }
   let data = {
     parentNode: elToReplace.parentNode,
-    replacement: widgetFrame,
+    replacement: shadowHost,
     origWidgetElem: elToReplace
   };
   if (widget.scriptSelectors) {
@@ -690,9 +693,9 @@ function createReplacementWidget(widget, elToReplace) {
         return;
       }
       e.preventDefault();
-      WIDGET_ELS[name] = WIDGET_ELS[name].filter(d => d.replacement != widgetFrame);
+      WIDGET_ELS[name] = WIDGET_ELS[name].filter(d => d.replacement != shadowHost);
       doNotReplace.add(elToReplace);
-      widgetFrame.replaceWith(elToReplace);
+      shadowHost.replaceWith(elToReplace);
     }, { once: true });
 
   }, false); // end of click handler
@@ -796,7 +799,7 @@ a:hover {
 
   widgetFrame.srcdoc = '<html><head><style>' + head_styles + '</style></head><body style="margin:0">' + widgetDiv.outerHTML + '</body></html>';
 
-  return widgetFrame;
+  return shadowHost;
 }
 
 /**
--- a/src/js/contentscripts/socialwidgets.js
+++ b/src/js/contentscripts/socialwidgets.js
@@ -396,10 +396,6 @@ function replaceSubsequentTrackerButtonsHelper(tracker_domain) {
   });
 }
 
-function _make_id(prefix) {
-  return prefix + "-" + Math.random().toString().replace(".", "");
-}
-
 function createReplacementWidget(widget, elToReplace) {
   if (!elToReplace.parentNode) {
     return null;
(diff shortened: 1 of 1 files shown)


src/js/contentscripts/socialwidgets.js
```diff
@@ -408,19 +404,22 @@ function createReplacementWidget(widget, elToReplace) {
   let name = widget.name;
 
   let widgetFrame = document.createElement('iframe');
+  let shadowHost = document.createElement("div");
+  let shadowRoot = shadowHost.attachShadow({ mode: "closed" });
+  shadowRoot.appendChild(widgetFrame);
```
lenacohen, line 409: I could replace the iframe with the shadowHost, but didn't see a benefit to doing that in the spirit of minimizing change. Let me know if you think I should.

tests/selenium/widgets_test.py
```diff
@@ -78,6 +78,12 @@ def set_up_widgets(self):
             "}(arguments[0]));"
         ), widgetsJson)
 
+    def switch_to_shadow_host_frame(self, selector):
+        shadow_host = self.find_el_by_css("div.widget-shadow-host")
```
ghostwords, line 82: Rather than exposing an identifier for tests, let's loop over all the divs and look for one with a shadow host that contains the iframe we want. (This code assumes the first matching div will be the one we want, which seems to work at the moment but may not in the future.)

Something like,

```python
    def switch_to_placeholder_frame(self, selector):
        for el in self.driver.find_elements(By.CSS_SELECTOR, "div"):
            try:
                shadow_root = el.shadow_root
            except NoSuchShadowRootException:
                continue
            iframe = shadow_root.find_element(By.CSS_SELECTOR, selector)
            self.driver.switch_to.frame(iframe)
            return
        raise NoSuchElementException("Failed to find placeholder frame")
```

src/js/contentscripts/socialwidgets.js
```diff
@@ -408,19 +404,24 @@ function createReplacementWidget(widget, elToReplace) {
   let name = widget.name;
 
   let widgetFrame = document.createElement('iframe');
+  let shadowHost = document.createElement("div");
+  shadowHost.className = "widget-shadow-host";
```
ghostwords, line 408: Let's remove this (see note below).

src/js/contentscripts/socialwidgets.js
```diff
@@ -408,19 +404,24 @@ function createReplacementWidget(widget, elToReplace) {
...
+  // Use an open shadow root on EFF test pages to allow our Selenium tests to access placeholder buttons
+  let shadowRoot = shadowHost.attachShadow({ mode: window.location.host === "efforg.github.io" ? "open" : "closed"});
```
ghostwords, line 410: We can make use of Selenium's shadow_root property instead.

ghostwords (commented):
Thank you! This round of feedback is about testing updates. I still have to test against real-world widget examples.

src/js/contentscripts/socialwidgets.js
lenacohen: Sounds good. I'll leave it as is for now.

tests/selenium/widgets_test.py
```diff
@@ -114,13 +120,13 @@ def assert_replacement(self, widget_name=None):
             widget_name = self.TYPE3_WIDGET_NAME
 
         try:
-            self.switch_to_frame(f'iframe[srcdoc*="{widget_name}"]')
-        except (StaleElementReferenceException, TimeoutException):
+            self.switch_to_shadow_host_frame(f'iframe[srcdoc*="{widget_name}"]')
+        except (StaleElementReferenceException, TimeoutException, NoSuchElementException):
```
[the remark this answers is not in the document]
lenacohen: Yes, in order for the tests to fail cleanly with the intended failure message when the placeholder frame isn't found.

tests/selenium/widgets_test.py
lenacohen: That works, thank you!

ghostwords:
Is there anything we can do for cases like https://www.twz.com/sea/ukraine-situation-report-russian-navy-creating-new-drone-regiments-in-wake-of-enemy-successes where our placeholder is newly hidden because it no longer overrides anything with a lower z-index?

Using shadow DOM isolates placeholder styles from page styles, which is great. Our z-index no longer working is less great though. While it's nice to no longer overlap navbars, I think it's a poor tradeoff if it means our placeholders can now be more easily hidden by site error messages.

PR #3069, opened 2025-03-08
Update ru/messages.json

wavy-cat:
This PR replaces a word with the correct variant in the Russian localization.


ghostwords (approved):
Thank you!

PR #3094, opened 2025-04-21
Update old-style FAQ links

ghostwords:
The space between the first placeholder and the text is for Safari.


lenacohen (approved):
Thank you!

Issue #3098, opened 2025-05-22
High CPU and memory usage and crashing in Chrome


Harmon758:
Using Version 2025.5.14 with Chrome Version 137.0.7151.41 on Windows 10, Privacy Badger is using up virtually all available CPU and memory off and on. It doesn't seem to be functional either, as clicking the toolbar icon results in the extension crashing.

When I check DevTools through the extension's service worker link, it seems to be continuously initializing/reloading every few seconds or so.

I've tried removing and re-adding the extension to no avail.

ghostwords:
Hello and thanks for opening an issue!

Are you able to install and use other content blocking extensions (uBlock Origin Lite, for example) in the same browser?

Could you try [installing Privacy Badger from unpacked source](https://github.com/EFForg/privacybadger/blob/master/doc/develop.md) in developer mode? I suggest trying the latest version from our [releases page](https://github.com/EFForg/privacybadger/releases) first, to see if you can reproduce the problem way. If you can, try loading the [previous version](https://github.com/EFForg/privacybadger/releases/tag/release-2025.3.27) next.

Let me know if you have any questions.

ghostwords:
Hmm, that's strange, the same code behaves differently depending on installation method.

You wrote that you tried uninstalling and reinstalling already, right? When you reinstall, do you get the "thanks for installing PB" tab after a few seconds, or does it never appear for whatever reason?

>When I check DevTools through the extension's service worker link, it seems to be continuously initializing/reloading every few seconds or so.

Could you post a screenshot of PB's service worker console output? I wonder if there are any hints there.

Harmon758:
> You wrote that you tried uninstalling and reinstalling already, right? When you reinstall, do you get the "thanks for installing PB" tab after a few seconds, or does it never appear for whatever reason?

Yes, I did before and that tab wasn't appearing.

Just to make sure, I just uninstalled and reinstalled it again and it still wasn't appearing. But I tried disabling and re-enabling it again and I think it was trying to load that page, but Chrome froze entirely. I ended Chrome's tasks in task manager and relaunched it, and that tab loaded properly, so I ended up with two "Thank you for installing Privacy Badger!" tabs open after I restored my tabs.

The extension seems to be working normally now, so I think it's resolved, although I'm not sure exactly what it was. Perhaps I needed to restart Chrome after I reinstalled it? I didn't do that the first time.

Apologies it seemed to be an issue with my local installation and not with the extension itself.
Thanks for helping me troubleshoot it.

PR #3100, opened 2025-05-30
Fix icon errors on service worker restart

ghostwords:
The errors seem to happen frequently in some Chrome environments when we provide more than five sizes to `setIcon()`.

When they happen, they seem to lead to the extension service worker failing to restart, and even Chrome crashes sometimes (when trying to open the popup).

#3098 is possibly related.

Following up on #3097, specifically https://github.com/EFForg/privacybadger/commit/1cccc95dbd6223b2e881a6e3524442dbda411fc5#diff-25daee2bd4a749c171c5bd0b87567f3fe306a1fdacb2a6d6785a3e1e6fd065f9


lenacohen (approved):
LGTM!

PR #3122, opened 2025-08-05
Fix tab order on tracking domains setting page and remove float layout

### LINKED ISSUE #3107: Wrong keyboard navigation order for toggles and "x" icons on options page

BeniRupp:
Previously the tab order in the list of tracking domains was: domain, x-Button, toggle.

Now, the order is as expected: domain, toggle, x-Button. This was achieved by changing the order of the DOM nodes inside a tracking domain row. The layout was then fixed by replacing the float layout with some grid and flex layouts to have a more robust layout with fewer magic numbers.

Fixes #3107

_This PR was part of the Hackergarten Stuttgart (Hack no. 427)._
_Co-authored by @itlinuxmaker_

### DIFF
--- a/src/js/htmlutils.js
+++ b/src/js/htmlutils.js
@@ -213,9 +213,11 @@ let htmlUtils = {
     <span class="ui-icon ui-icon-alert tooltip breakage-warning" title="${breakage_warning_tooltip}"></span>
     <span class="origin-inner tooltip" title="${domain_tooltip}">${dnt_html}${shield_icon}${fqdn}</span>
   </div>
+  <div class="toggle-container">
+    ${htmlUtils.getToggleHtml(fqdn, action, blockedFpScripts)}
+    <a href="" class="honeybadgerPowered tooltip" title="${undo_arrow_tooltip}"></a>
+  </div>
   <a href="" class="removeOrigin">&#10006</a>
-  ${htmlUtils.getToggleHtml(fqdn, action, blockedFpScripts)}
-  <a href="" class="honeybadgerPowered tooltip" title="${undo_arrow_tooltip}"></a>
 </div>
       `.trim();
     };
--- a/src/js/options.js
+++ b/src/js/options.js
@@ -646,7 +646,7 @@ function getOriginAction(domain) {
 function revertDomainControl(event) {
   event.preventDefault();
 
-  let domain = $(event.target).parent().data('origin');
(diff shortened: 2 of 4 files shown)


ghostwords:
Thank you!! I will be able to review your contribution mid next week.

ghostwords:
A couple of our [functional tests](https://github.com/EFForg/privacybadger/blob/master/doc/tests.md#functional-tests) are failing. I wonder if it's because the tests need to be updated (probably because the selectors changed) or if they flagged a regression.

BeniRupp:
Oh.. yes, that could be a problem with the selectors. I can have a look on the tests next week.

lenacohen:
Thank you for working on this! 

I noticed that your changes caused the sliders and their icons to shift left. We want to keep the layout of the popup the same.
**Before:**
Screenshot 2025-08-08 at 5 23 38 PM


**After:**
Screenshot 2025-08-08 at 5 23 49 PM


On the options page, the focused domain should have the default highlight styling, rather than this orange box:
Screenshot 2025-08-08 at 5 12 56 PM

lenacohen:
Let us know if you have any questions about our feedback. Thanks again for contributing!

itlinuxmaker:
@lenacohen _"I noticed that your changes caused the sliders and their icons to shift left. We want to keep the layout of the popup the same."_ 
I'm going to jump into the discussion here. Firstly, because I've also been pondering a possible solution, and secondly, because I think the whole thing is a disaster.
So you want the layout to be consistent, and so do we. Secondly, the functionality regarding keyboard operation should also be guaranteed. We've both solved that – perhaps at the expense of the layout. But the problem itself is solved.
If there are still requests regarding the layout, it's your turn.
The catastrophic thing is the fact that software is being released and expected to be used by less technically savvy users that hasn't been sufficiently and thoroughly tested. If that had been the case, this issue wouldn't exist at all.
So please be grateful that you were helped instead of making additional demands now.

ghostwords:
@itlinuxmaker Providing, accepting and acting on feedback is a standard part of software development. Furthermore, please note that this project is governed by the [EFF Public Projects Code of Conduct](https://www.eff.org/pages/eppcode). Specifically:

>Please be kind and courteous. Please take care not to be, or appear to be, mean or rude. Keeping this project a welcoming place is a shared responsibility involving each of us.

Issue #3138, opened 2025-09-08
Tracker list icons missing alt-text or tooltip descriptions


lenacohen:
Several icons that appear in the expanded tracker list in Privacy Badger's popup are not screanreader accessible. They are missing alt-text and the tooltips that appear on hover are not read by screenreaders. A possible fix is using the tooltip language as alt-text for the icons.

The inaccessible icons are:

- The arrow to return control to Privacy Badger

<img width="319" height="98" alt="Image" src="https://github.com/user-attachments/assets/77e8ae34-8875-4921-a7f2-274c69605897" />

- The icons explaining what each slider option does

<img width="271" height="90" alt="Image" src="https://github.com/user-attachments/assets/3611ef72-4ac1-4d5b-99a4-5a12dc1adb13" />

- The DNT icon 

<img width="261" height="69" alt="Image" src="https://github.com/user-attachments/assets/5aecfeb1-8205-420d-80ab-f17022089b2e" />

Issue #3131, opened 2025-08-27
The "to enable logging into websites with your Google account" tooltip may not be screen reader accessible


ghostwords:
Let's audit #2861 with screen reader.

Issue #3133, opened 2025-08-27
The Tracking Domains tab on options page may freeze the browser when used with screen reader


ghostwords:
The Tracking Domains tab is pretty slow as is, but it may be worse for screen reader users.

Let's audit the Tracking Domains tab with screen reader, and see what we can do to avoid the following type of browser error:

<img width="698" height="403" alt="Image" src="https://github.com/user-attachments/assets/8ceaa25d-66e5-4924-ab64-646eebf44353" />

PR #3143, opened 2025-09-19
Make tooltip content in popup screenreader-accessible

lenacohen:
**Addresses:**
- https://github.com/EFForg/privacybadger/issues/3138
- https://github.com/EFForg/privacybadger/issues/3132
- https://github.com/EFForg/privacybadger/issues/3131
- Made the collapsed/expanded state of the tracker list screenreader-accessible
- Added `role="radiogroup"` and an aria-label to the slider group to provide screenreaders with more context for slider buttons

Tested with the VoiceOver screenreader on MacOS

**Test instructions:**
- Navigate to each of the elements that have a tooltip:
  - EFF logo
  -  "trackers" link
  - Icons above the tracker list representing the slider options
  - Domains in the tracker list
  -  Domain slider options
  - Breakage warning tooltip
  - "Click to return control" tooltip
  - Google sign-in warning tooltip
  - DNT icon
- To see the breakage warning tooltip, move a slider in the yellow (cookie-block) position to the red (block) position
Screenshot 2025-09-19 at 5 00 06 PM

- To see the "Click to return control" tooltip, move a slider to any position
Screenshot 2025-09-19 at 4 58 03 PM

- To see the "DNT" icon, visit https://www.nytimes.com/ and scroll down the tracker list to the domain with the DNT icon
Screenshot 2025-09-19 at 5 02 37 PM

- To see the Google sign-in warning tooltip, visit https://stackoverflow.com/ or another site with Google sign-in
 Screenshot 2025-09-19 at 4 56 31 PM

### DIFF
--- a/src/_locales/en_US/messages.json
+++ b/src/_locales/en_US/messages.json
@@ -115,6 +115,16 @@
         "message": "Move the slider right to allow a domain",
         "description": "Domain slider list header tooltip."
     },
+    "popup_slider_label": {
+        "message": "Blocking options for $DOMAIN$",
+        "description": "Screenreader-accessible label for domain sliders",
+        "placeholders": {
+            "domain": {
+                "content": "$1",
+                "example": "example.com"
+            }
+        }
+    },
     "badger_status_block": {
         "message": "Blocked $DOMAIN$",
         "description": "Tooltip shown over a domain name with a red slider in the domain slider list. This is a status (\"example.com was blocked\"), not an action (\"click here to block example.com\").",
--- a/src/js/htmlutils.js
+++ b/src/js/htmlutils.js
@@ -126,7 +126,7 @@ let htmlUtils = {
       let id = fqdn.replace(/\./g, '-');
       return `
 <div class="switch-container ${action}">
-  <div class="switch-toggle switch-3 switch-candy">
+  <div class="switch-toggle switch-3 switch-candy" role="radiogroup" aria-label="${i18n.getMessage('popup_slider_label', fqdn)}">
     <input id="block-${id}" name="${fqdn}" value="${constants.BLOCK}" type="radio" aria-label="${tooltips.block}" ${is_checked(constants.BLOCK, action)}>
     <label title="${tooltips.block}" class="tooltip" for="block-${id}"></label>
     <input id="cookieblock-${id}" name="${fqdn}" value="${constants.COOKIEBLOCK}" type="radio" aria-label="${tooltips.cookieblock}" ${is_checked(constants.COOKIEBLOCK, action)}>
--- a/src/js/htmlutils.js
+++ b/src/js/htmlutils.js
@@ -150,7 +150,7 @@ let htmlUtils = {
     return function () {
       return `
       <div class="dnt-compliant">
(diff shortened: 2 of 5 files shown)


src/js/htmlutils.js
```diff
@@ -150,7 +150,7 @@ let htmlUtils = {
     return function () {
       return `
       <div class="dnt-compliant">
-        <a target=_blank href="https://privacybadger.org/#-I-am-an-online-advertising-tracking-company.--How-do-I-stop-Privacy-Badger-from-blocking-me"><img src="${dnt_icon_url}"></a>
+        <a target=_blank href="https://privacybadger.org/#-I-am-an-online-advertising-tracking-company.--How-do-I-stop-Privacy-Badger-from-blocking-me" aria-label="${i18n.getMessage('dnt_tooltip')}"><img src="${dnt_icon_url}" alt=""></a>
```
ghostwords, line 153: Let's cache this message here too, as we already cache `dnt_icon_url`.

src/js/popup.js
```diff
@@ -571,6 +573,11 @@ function createBreakageNote(domain, i18n_message_key) {
     autoClose: false,
     content: chrome.i18n.getMessage(i18n_message_key),
     functionReady: function (tooltip) {
+      // Add aria-hidden to ensure that the tooltip isn't read twice by screenreaders
+      // It should be read along with the selected slider option (via aria-describedby), and not where it's inserted at the end of the popup DOM
+      $(tooltip.elementTooltip()).attr({role: "tooltip", "aria-hidden": "true"});
+      $slider_allow .closest(".switch-toggle").find("input:checked").attr("aria-describedby", tooltip.elementTooltip().id);
```
ghostwords, line 579: Nit: space before the dot.

src/_locales/en_US/messages.json
```diff
@@ -115,6 +115,16 @@
         "message": "Move the slider right to allow a domain",
         "description": "Domain slider list header tooltip."
     },
+    "popup_slider_label": {
+        "message": "Blocking options for $DOMAIN$",
```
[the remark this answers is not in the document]
lenacohen: What about "Settings for $DOMAIN$" or "Blocking mode for $DOMAIN$"?

Other blocking extensions don't have sliders (except for UBlock's "filtering mode") and/or are not very screenreader accessible

lenacohen:
> We want want to have the same stuff happen in the options page. If we don't do it here, we might want to have the helper functions live in a common location so that we can do it on the options page when ready.
> 
> It doesn't look like the "Sign in with google" breakage warning is accounted for when keyboard navigating in this PR.

I think we should wait to update the domain tooltips on the option page until https://github.com/EFForg/privacybadger/issues/3133 is fixed (although I haven't been able to replicate on my computer). I moved the helper functions into a common location so that it's easy when the option page is ready. 

Do you mean that the "Sign in with google" warning should be tabbable? I added `aria-hidden` so it's read as the description of the selected toggle, but not read again at the end of the DOM. I'll [need to remove](https://dequeuniversity.com/rules/axe/4.4/aria-hidden-focus) that if we make it tabbable. Separately, it may be unclear to keyboard and non-keyboard users that clicking on the tooltip closes it, since I don't think that's a common pattern (unlike click-outside-to-close). To make it more accessible, we may want to add an X button and make that tabbable.

ghostwords:
>Do you mean that the "Sign in with google" warning should be tabbable?

Yeah, exactly. While tabbing through the popup, the keyboard never focuses on the warning, so you can't dismiss it with the keyboard.

PR #3144, opened 2025-09-29
Make GPC/DNT exceptions remotely configurable

ghostwords:
I [already deployed](https://www.eff.org/files/pbconfig.json) the `gpc` subkey so you don't have to disable config updates from remote when testing.


src/data/pbconfig.json
```diff
@@ -9,6 +9,11 @@
         "DNT Policy v1.0 no-eof-newline": "7861462d500fbb6ccb74c614782f4937377bec76"
     },
     "sitefixes": {
+        "gpc": [
+            "www.costco.com",
+            "www.delta.com",
+            "fivethirtyeight.com"
```
lenacohen, line 15: I think we can remove, since this site got shut down and now redirects to https://abcnews.go.com/politics

PR #3146, opened 2025-10-22
Re-implement breakage note with close button and without tooltipster

### LINKED ISSUE #3131: The "to enable logging into websites with your Google account" tooltip may not be screen reader accessible

lenacohen:
- Re-implement the Google sign-in breakage note to move it from the end of the DOM to a logical place for screenreader users 
  - Tooltipster does not have an option for this and seemed to undo moving the tooltip root in functionReady
- Made closing the tooltip more clear/accessible by removing click to close and adding a close button (with slightly different styling than https://github.com/EFForg/privacybadger/pull/3034). This also allows users to highlight/copy text without it closing the tooltip.

Fixes #3131

Closes #3034

Follows up on #2861

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -618,18 +618,22 @@ function createBreakageNote(domain, i18n_message_key) {
   // Collect originally focusable elements we suppress
   let suppressed = [];
   function suppressOverlapping() {
-    // Ensure that only visible elements can receive keyboard focus
+    // Ensure that obscured elements will not receive keyboard focus
     // (https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html)
     let tooltipRect = $tooltip[0].getBoundingClientRect();
-    document.querySelectorAll(htmlUtils.focusableSelectors).forEach(el => {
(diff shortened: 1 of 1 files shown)


lenacohen:
Current implementation:
Screenshot 2025-10-22 at 10 46 17 AM

Slightly different margin and placement for the close button (with existing tooltip shown for comparison):
Screenshot 2025-10-21 at 12 53 25 PM

Screenshot 2025-10-21 at 12 50 30 PM

ghostwords (commented):
The sign-in note may take up more header space, depending on the language. In English, the note covers the show/hide sliders header (with the "what is a tracker" link), at least on sites with enough domains like `nytimes.com`. In Russian, the note will also cover most of the Badger logo and top right corner icons.

I don't think we want to be able to tab through things obscured by the note. I also think the note should be the first thing we tab to, followed by the domain toggle the note points to.

Maybe showing the note should gray out header elements and remove them from tab order? (I'm not sure how to solve this yet, but it seems like this is a problem.)

Issue #3103, opened 2025-06-05
Domain toggles are always LTR even in RTL locales


ghostwords:
In the popup and on the Tracking Domains tab on the options page.

![Image](https://github.com/user-attachments/assets/ca1593c0-cc7e-44b1-8d12-0a5a894ed562)

Fixing this probably means replacing floats with a modern (flex?) layout.

Related to #3107.

giant1izards:
Just to clarify, is the ask for the toggle switches to be green-yellow-red in a RTL locale, and red-yellow-green in a LTR locale? It was pretty simple to switch the higher-level layout of the page (i.e. have the x on the left side and the FQDN on the right), but after some exploratory efforts I think making the toggle switch itself be reversible would be a significant lift. Like, essentially recreating the whole component from scratch.

PR #3152, opened 2025-11-21
fix: change tab order on tracking domains page

### LINKED ISSUE #3107: Wrong keyboard navigation order for toggles and "x" icons on options page

giant1izards:
This fixes #3107 

I used [PR #3122](https://github.com/EFForg/privacybadger/pull/3122) as a starting point, and made sure the tests pass and the UI was not affected (too badly). Used flexbox on the tracking domain rows so that tabbing works implicitly with the HTML elements.

Attempted to knock out #3103 also since they are related, but I wasn't able to find a quick, clean way to make the toggle-switch direction-agnostic so I'll leave that for another day.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,6 @@ tests/.pytest_cache
 
 pkg
 release-utils/config.sh
+
+# development tooling
+.devcontainer/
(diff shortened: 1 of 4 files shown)


giant1izards:
Thanks for the feedback, I will work on cleaning those things up. I think for the alignment issues I will see if using CSS grid instead of flexbox helps, it should be easier to keep things in structured columns. Shouldn't be much trouble.

The RTL mode things also make sense.

ghostwords (commented):
Nice, this seems close to a solution for #3107.

It looks like the "x"s are no longer aligned to the toggles and domain names the way they are now.

Let me know if anything is unclear and thanks again for working on this.

lenacohen (commented):
Thank you for contributing to Privacy Badger! In addition to Alexei's feedback, could you double check if your CSS changes also need to be applied inside media queries for different screen sizes?

PR #3191, opened 2026-06-11
Fix header font issues in some locales on welcome page

### LINKED ISSUE #3190: Header ChunkFive font looks bad in some locales on welcome page
Maybe in other places too? The screenshots below are from Chrome but the same problem exists (and looks even worse somehow) in Safari.

Image
Image

lenacohen:
- Switch welcome page header fonts from ChunkFive to NotoSans
- Fixes #3190


ghostwords (approved):
Thank you!

PR #3195, opened 2026-06-25
fix: prevent TypeErrors when optional extension APIs are unavailable

MatrixNeoKozak:
This change introduces safety guards around calls to the optional `chrome.cookies` and `chrome.browserAction` APIs. In environments where these APIs are unavailable (such as in private window contexts, environments without cookie permissions, or mock/test setups), calling their methods or using utility functions like `utils.hasOwn` directly on them resulted in unexpected `TypeError` exceptions that crashed initialization and request processing.

Specifically:
1. Checked for the existence of `chrome.cookies` in `testCookiesFirstPartyDomain()` inside `src/js/background.js` before attempting to invoke `chrome.cookies.getAll`.
2. Safely guarded the `chrome.browserAction` badge text color modification during `Badger` storage ready initialization in `src/js/background.js` to ensure the extension starts up gracefully if `chrome.browserAction` is not defined.
3. Added an early exit check in `checkForPixelCookieSharing()` within `src/js/heuristicblocking.js` to verify `chrome.cookies` exists before starting pixel cookie share verification.

### DIFF
--- a/src/js/background.js
+++ b/src/js/background.js
@@ -43,6 +43,9 @@ import utils from "./utils.js";
  * firstPartyDomain is not currently supported in Chrome.
  */
 function testCookiesFirstPartyDomain() {
+  if (!chrome.cookies) {
+    return false;
+  }
   try {
     chrome.cookies.getAll({
       firstPartyDomain: null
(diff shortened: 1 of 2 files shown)


ghostwords:
Hello and thanks for opening a pull request!

Could you talk a little about what prompted you to look into this? Did you notice errors somewhere and decide to look into them, or? Which browser do you use?

These code paths belong to functionality that we restrict to the developer environment. In other words, we never release this code to Privacy Badger users.

MatrixNeoKozak:
Thanks for the context! You're right — I missed that these code paths are restricted to the developer environment and that `testCookiesFirstPartyDomain` already wraps the call in a try/catch. This PR doesn't address a real user-facing issue, so I'll close it. Appreciate you taking the time to explain.

PR #3200, opened 2026-06-30
Skip injectScript() on CSP-restricted pages

runephilosof:
Pages with a restrictive Content-Security-Policy block inline script injection via injectScript(), which pollutes the browser console and the Errors button on chrome://extensions/ without providing any benefit.

Fix this by reading the Content-Security-Policy response header in onHeadersReceived() and recording whether inline scripts are allowed on the frame data object. The four message handlers that gate features relying on injectScript() — detectFingerprinting, detectSupercookies, checkClobberingEnabled, and checkDNT — now return false when the frame's CSP blocks inline scripts, so injectScript() is never called.

CSP parsing handles:
- script-src falling back to default-src
- 'unsafe-inline' (allows injection)
- nonces and hashes (make 'unsafe-inline' ineffective, per CSP3)
- 'strict-dynamic' (also makes 'unsafe-inline' ineffective)
- multiple CSP headers (all must permit inline scripts)

Add QUnit unit tests for the CSP parsing logic.
Add Selenium test for CSP-restricted injectScript() suppression

The @pytest.mark.flaky(reruns=5) decorator matches the pattern used by other Chrome/Edge tests in this suite that share a known BiDi timing issue on first browser startup.

Related: #3053

### DIFF
--- a/src/js/webrequest.js
+++ b/src/js/webrequest.js
@@ -476,6 +476,45 @@ function onBeforeSendHeaders(details) {
 }
 
 /**
+ * Checks whether the page's Content-Security-Policy allows inline scripts.
+ *
+ * Returns true when inline scripts are allowed (or when there is no
+ * relevant CSP directive), false when they are blocked.
+ *
+ * Handles multiple CSP headers (all must permit inline scripts for the
+ * result to be true). Treats 'unsafe-inline' as ineffective when nonces
+ * or hashes are also present (CSP Level 3 behaviour).
+ *
+ * @param {Array<{name: string, value: string}>} headers
+ * @returns {Boolean}
+ */
+function inlineScriptsAllowedByCsp(headers) {
+  for (const header of headers) {
+    if (header.name.toLowerCase() !== 'content-security-policy') {
+      continue;
+    }
+    const policy = header.value || '';
+    const scriptSrcMatch = /(?:^|;)\s*script-src\s+([^;]+)/i.exec(policy);
+    const defaultSrcMatch = /(?:^|;)\s*default-src\s+([^;]+)/i.exec(policy);
+    const directive = scriptSrcMatch ? scriptSrcMatch[1] : (defaultSrcMatch ? defaultSrcMatch[1] : null);
+    if (!directive) {
+      continue;
+    }
+    if (/'unsafe-inline'/i.test(directive)) {
+      // In CSP3, 'unsafe-inline' is ignored when nonces, hashes, or
+      // 'strict-dynamic' are present
+      if (/(?:nonce|sha(?:256|384|512))-/i.test(directive) ||
+        /'strict-dynamic'/i.test(directive)) {
+        return false;
+      }
+    } else {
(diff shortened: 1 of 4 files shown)


runephilosof:
CSP is a defence-in-depth protection.
To be useful, CSP violation reports should only show actual sanitization problems in the app it is protecting.
Extension noise hides attackers' unsuccessful attempts.

ghostwords:
Hello and thanks for your interest in contributing to Privacy Badger!

Could you review and respond here to the applicable points under the "Policy regarding AI-generated contributions" subheading on the [Coding with EFF](https://www.eff.org/about/opportunities/volunteer/coding-with-eff) page?

runephilosof:
I have used AI assistance, I have reviewed, understood, and adjusted it.

runephilosof:
An alternative approach would be to simply inject the necessary `sha-*` into the CSP, or when there is a `nonce-*` in it, record it and use it where we inject scripts.

That way the extension would still work.

ghostwords:
Thanks for clarifying.

My sense is that this approach partially (what about sites with CSP in meta tags) works around a nice-to-fix (not urgent) problem that we already mean to address properly when the extension APIs to do so become available. It's not clear to me that the tradeoff of reviewing and maintaining this additional complexity is worth it.

Could you talk about what led you to look into this in the first place?

runephilosof:
CSP violation reports are supposed to help you
1. add directives to your CSP.
2. warn you that an attacker was able to execute javascript on your site, so you can fix that before they might figure out other attacks.

Noise from extensions make that really hard.
So when I noticed that I had an extension installed, which was one of the culprits making noise, I wanted to improve it.

So, it might be "nice to fix" from the extensions point of view.
However, it is quite problematic for web sites and the adoption of secure CSP settings that the extension noise is reduced.

This PR changes no user visible functionality for the extension, because CSP is already stopping that functionality.
Support for meta tags might be possible in a follow up.

If you prefer, I can implement the alternative approach instead, which would fix extension functionality on sites with secure CSP.

ghostwords:
I don't think we should attempt to patch site CSPs as a workaround because the failure scenario of undoing CSP protections is very bad.

ghostwords:
>So when I noticed that I had an extension installed, which was one of the culprits making noise

This is what I was asking about. Do you run a website and did you notice this in your website logs?

I am sorry PB contributes to log noise, but we do plan to fix this issue as soon as an appropriate solution is available.

ghostwords:
As I wrote back in https://github.com/EFForg/privacybadger/issues/3053#issuecomment-2651844011,  `dom.execute()` (synchronous cross-world content script injection) and `dom.createPort()` (secure communication) appear to be the way forward that solve this and other issues (injection timing problems, insecure main world-to-content script communication) in one stroke. The proposals appear to be stalled out at this point, however.

ghostwords:
>Do you run a website and did you notice this in your website logs?

Also, which browsers do you notice these reports coming from?

PR #3202, opened 2026-07-09
Update license name from GPL v3+ to GPL v3 or later

thomasboom:
I updated the license name used in the README.md and the LICENSE file to use the license name GPL v3 or later instead of GPL v3+, a clearer, more universally agreed on name. This does not change the license itself.


ghostwords:
Thank you!

Issue #3141, opened 2025-09-09
Keyboard navigation not properly trapped within welcome page reminder/critical error/important update notification modal


ghostwords:
The following modal overlays do not [properly trap keyboard navigation](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/):

https://github.com/EFForg/privacybadger-mv3/blob/9bf7e13f5b29765c665d989b5666397e3ef21ca6/src/skin/popup.html#L36-L73

To see the welcome page reminder modal overlay (`div#instruction-text`):

1) Add PB to a browser
2) Wait for the welcome page to appear
3) Close the welcome page
4) Navigate to any website, such as `example.com`
5) Open PB's popup (by clicking on PB's icon, either on the browser toolbar, or inside the "puzzle piece" icon on the browser toolbar)

Once the overlay is open, navigating with <kbd>Tab</kbd> should cycle between the close icon and the "CTA" (call to action) button.

The bug is that continuous tabbing leaves the overlay and continues into the rest of the UI (grayed out and partially obscured by the modal).

<img width="558" height="409" alt="Image" src="https://github.com/user-attachments/assets/638314e5-7759-47c9-9272-1b20873d7a31" />

This follows up on #3140.

Related to #3105.

PR #3206, opened 2026-07-15
Fix: Trap focus in overlaid popups (Issue #3141)

Capezzuto:
This code update should address issue #3141 (Keyboard navigation not properly trapped within welcome page reminder/critical error/important update notification modal). Changes are scoped to elements in "#instruction" dialog element in popup.html.

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -67,6 +67,7 @@ function showNagMaybe() {
   var $nag = $("#instruction");
   var $outer = $("#instruction-outer");
   let intro_page_url = chrome.runtime.getURL("/skin/firstRun.html");
+  let $nagFocusables = $("#instruction").find(htmlUtils.focusableSelectors);
 
   function _setSeenComic(cb) {
     chrome.runtime.sendMessage({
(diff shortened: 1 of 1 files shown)


ghostwords:
Thank you! We will take a look in the next couple of days.

src/js/popup.js
```diff
@@ -81,30 +82,66 @@ function showNagMaybe() {
...
+      return el.id === "fittslaw" || $(id).has(el).length;
+    });
+    let $firstFocusableEl = $currentNagFocusables.first();
+    let $lastFocusableEl = $currentNagFocusables.last();
+
+    return function(e) {
+      let is_tab = e.key === "Tab" || e.keycode === 9;
+      if (!is_tab) {
+        return;
+      }
```
lenacohen, line 97: `keycode` will be undefined because the property is `keyCode`, but I think you can remove the check because the property is deprecated (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and `key` is supported by Privacy Badger's minimum browser versions

lenacohen (commented):
Thank you for opening this PR! I tested each of the nag modals and focus trapping works! 

However, the trap works now because `#instruction` comes before all other focusable elements in the DOM. Focus isn't explicitly moved into the modal when it opens. If a focusable element is ever added earlier in `popup.html`, the first Tab press will focus there instead of inside the modal, and focus won't be trapped in the modal until the user tabs into it.

ghostwords:
Thank you for contributing to Privacy Badger!

If we take a step back, what we're dealing with here is a partially-implemented `dialog` element. This pull request is a step towards a more complete implementation, one that correctly handles keyboard navigation.

As far as I can see though, there is no need at this point to re-implement `dialog`. I think we should switch our implementation to `dialog`, which will automatically take care of keyboard trapping, as well as of other things that we may or may not be doing properly in our custom implementation.

Would you be interested in opening a new pull request that fixes #3141 by switching `#instruction` from `div` to `dialog`?

Capezzuto:
Sure, I work on that approach. With that being the case, please feel free to close this PR.

ghostwords:
Let me know if I can help with the `dialog` pull request! Feel free to open a work-in-progress version that we can complete together.

Capezzuto:
@ghostwords Just wanted to clarify something...In this comment, did you mean that we should switch the implementation to the jQuery UI `dialog` widget to handle focus-trapping?

> As far as I can see though, there is no need at this point to re-implement `dialog`. I think we should switch our implementation to `dialog`, which will automatically take care of keyboard trapping, as well as of other things that we may or may not be doing properly in our custom implementation.

ghostwords:
I meant we should switch to the native `dialog` element:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog
- https://webkit.org/blog/12209/introducing-the-dialog-element/
- https://css-tricks.com/some-hands-on-with-the-html-dialog-element/

PR #3207, opened 2026-07-16
Address accessibility feedback

lenacohen:
- Added alt text to the EFF logo on the welcome page
- Added the tracker domain to the "Remove selected" button labels on the Tracking Domains tab of the options page 
- Increased the size of text on orange buttons to meet WCAG standards for color contrast 
- Updated the Privacy Policy text link color to meet WCAG standards for color contrast (in dark mode)

### DIFF
--- a/src/js/htmlutils.js
+++ b/src/js/htmlutils.js
@@ -176,7 +176,8 @@ let htmlUtils = {
   // TODO origin --> domain/FQDN
   getOriginHtml: (function () {
 
-    const breakage_warning_tooltip = i18n.getMessage('breakage_warning_tooltip');
+    const breakage_warning_tooltip = i18n.getMessage('breakage_warning_tooltip'),
+      remove_button_label = i18n.getMessage('options_domain_remove', 'XXX');
 
     return function (fqdn, action, show_breakage_warning, show_breakage_note, blockedFpScripts) {
       action = escape_html(action);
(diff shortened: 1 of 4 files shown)


src/js/htmlutils.js
```diff
@@ -213,6 +213,8 @@ let htmlUtils = {
       // construct HTML for domain
       let domain_tooltip = htmlUtils.getActionDescription(action, fqdn, blockedFpScripts);
 
+      let remove_button_label = `${remove_button_intro}: ${fqdn}`;
```
ghostwords: I don't think we can assume this will read well in all languages.

Also, "Remove selected" is from the disabled sites form and reusing it here was a hack.

I think we should add a new message.

src/skin/css/firstRun.css
```diff
@@ -798,7 +801,7 @@ body {
     color-scheme: dark;
   }
 
-  body, #pb-privacy-policy {
+  body, #pb-privacy-policy a {
```
ghostwords: Is `color: #707070` not too low contrast in light mode also?

ghostwords: We should underline the privacy policy link to make it clear it's a link, same as with "EFF" in the header.

lenacohen: It wasn't too low contrast for WCAG level AA, but I'll make it black to meet AAA standards! I also added made the orange links black to meet color contrast requirements and added underlines to make it clear that they're links.

src/js/htmlutils.js
```diff
@@ -213,6 +212,8 @@ let htmlUtils = {
       // construct HTML for domain
       let domain_tooltip = htmlUtils.getActionDescription(action, fqdn, blockedFpScripts);
 
+      let remove_button_label = i18n.getMessage('options_domain_remove', fqdn);
```
ghostwords: Calls to `chrome.i18n.getMessage()` are "heavy" and should be cached in places that will be repeatedly invoked.

lenacohen: Ah yes I forgot about that, fixed!

ghostwords (approved):
Thank you!

PR #3209, opened 2026-07-19
Disposable Chrome profiles in test runs

runephilosof:
instead of the persistent default profile, for isolation between runs skip the First Run interstitial/checks a fresh profile would otherwise show, which can delay/break the initial navigation

Some Linux distros (e.g. Debian/PureOS) ship a system-installed "privacy-badger" package that Chromium auto-loads as an external extension into every profile, regardless of --user-data-dir. This results in a second, unrelated Privacy Badger instance (with its own extension ID) running alongside the copy under test, e.g. an extra firstRun.html tab that our welcome page cleanup code (see run() below) can't recognize/close, since it compares against chrome.runtime.getURL() of the *tested* extension's ID. Explicitly allowlisting only our own extension prevents this.

### DIFF
--- a/tests/selenium/pbtest.py
+++ b/tests/selenium/pbtest.py
@@ -8,7 +8,7 @@ import time
 import unittest
 
 from contextlib import contextmanager
-from shutil import copytree, which
+from shutil import copytree, rmtree, which
 
 from selenium import webdriver
 from selenium.common.exceptions import (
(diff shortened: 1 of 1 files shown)


ghostwords:
>instead of the persistent default profile

We already get a new browser profile:

>By default, ChromeDriver creates a new temporary profile for each session.

-- https://developer.chrome.com/docs/chromedriver/capabilities

>for isolation between runs skip the First Run interstitial/checks a fresh profile would otherwise show, which can delay/break the initial navigation

We already don't get these.

runephilosof:
Sorry anout that.
I created it because I saw problems from a globally installed extension. So maybe only the 

--disable-extensions-except

ghostwords:
It is a poor use of our time to try to figure out which portion of your auto generated contributions are rooted in reality and which come from a convincing text generation machine.

Do you see that the text generators have no concept of meaning or understanding? They are very good however at wasting our time.

If there is a real issue, please open a bug report without making use of text generation tools.

PR #3210, opened 2026-07-19
Fix/driver quit

runephilosof:


ghostwords:
If all retries fail and `driver` remains `None`, we have a bigger problem than an AttributeError.

runephilosof:
Indeed. But I also think that the small changes here are an obvious improvement that will make the error report more easy to read.

PR #3211, opened 2026-07-19
Stabilize selenium

runephilosof:
Chrome/Edge tests sometimes started interacting with the extension
before its background/runtime was actually ready, causing flaky
failures.


ghostwords:
If this is an actual issue, please first open a bug report with clear reproduction steps.

PR #3215, opened 2026-07-21
Add welcome page images for right-to-left locales

lenacohen:
- Update welcome page images for RTL
- Switch to new version of pinning GIF that fixes slight misalignment (only visible in dark mode)


ghostwords (approved):
Thank you!

PR #3217, opened 2026-07-23
Fix YouTube video link for non-JS embeds

ghostwords:
Our video link for YouTube iframe placeholders currently results in an "Error 153 / Video player configuration error" page, forcing an extra click to get to the actual video page. Sometimes getting to the actual video page is fully broken, [for example](https://www.youtube.com/embed/H4dGpz6cnHo?list=PLVAh-MgDVqvDUEq6qDXqORBioE4Yhol_z) (from https://aftermath.site/backrooms-movie-parsons-youtube/).

This fix currently passes on all existing parameters, including junk like "&feature=oembed". It seems better to pass on too much (doesn't do anything) than too little (some expected functionality is missing).

Video ID validation is currently limited to checking length (eleven characters) and that it isn't "videoseries" (a variant list embed type that happens to be eleven characters long). Are there any other 11-character sequences that we should exclude? As far as video ID validation goes, it's better to have false negatives than false positives.

Could add a special case to handle properly generating a video link for "/embed/videoseries" embeds.

This follows up on 23bb32498, which was the same fix but for YT API-inserted embeds.

### DIFF
--- a/src/js/contentscripts/socialwidgets.js
+++ b/src/js/contentscripts/socialwidgets.js
@@ -499,7 +499,15 @@ function createReplacementWidget(widget, elToReplace) {
     // use the frame URL for framed widgets
     if (elToReplace.src) {
       widget_url = elToReplace.src;
-      if (widget_url.startsWith("https://embed.bsky.app/embed/")) {
+      if (widget.name == "YouTube") {
+        let yuri = (new URL(widget_url)),
+          ypath = yuri.pathname.slice(1).split("/");
+        if (ypath.length == 2 && ypath[0] == "embed" &&
+            ypath[1].length == 11 && ypath[1] != "videoseries") {
+          let ysearch = (yuri.search ? "&" + yuri.search.slice(1) : "");
+          widget_url = "https://www.youtube.com/watch?v=" + ypath[1] + ysearch;
+        }
+      } else if (widget_url.startsWith("https://embed.bsky.app/embed/")) {
         // Bluesky
         let buri = (new URL(widget_url)).pathname.split("/");
         if (buri[2] && buri[2].startsWith("did:") && buri[4]) {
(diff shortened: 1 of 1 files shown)


lenacohen:
Code LGTM and works as expected on sites with this type of embed (i.e. https://brand-innovators.com/watch-every-super-bowl-ad-2/, https://www.foxsports.com/stories/nfl/best-super-bowl-commercials-2026, https://artofthemovies.co.uk/blogs/original-movie-posters/our-coming-attraction-18-of-the-greatest-movie-trailers-ever-made).

As expected, it doesn't work on sites with "/embed/videoseries" embeds (i.e. https://routenote.com/blog/top-playlists-youtube-music/, https://www.musicbloggersnetwork.com/watch/, https://forums.insertcredit.com/t/highly-specific-music-playlists-on-youtube-or-what-have-you/5252/14). We could get the link to go to the playlist that the embed is part of by adding this branch, but I'm not sure how to link to the first video in the playlist (like the embeds seem to do):
```
     else if (ypath.length == 2 && ypath[1] == "videoseries") {
          let playlist_id = yuri.searchParams.get("list");
          if (playlist_id) {
            let params = new URLSearchParams(yuri.search);
            params.set("list", playlist_id);
            widget_url = "https://www.youtube.com/playlist?" + params.toString();
          }
        }
```

src/js/contentscripts/socialwidgets.js
```diff
@@ -499,7 +499,15 @@ function createReplacementWidget(widget, elToReplace) {
     // use the frame URL for framed widgets
     if (elToReplace.src) {
       widget_url = elToReplace.src;
-      if (widget_url.startsWith("https://embed.bsky.app/embed/")) {
+      if (widget.name == "YouTube") {
+        let yuri = (new URL(widget_url)),
+          ypath = yuri.pathname.slice(1).split("/");
+        if (ypath.length == 2 && ypath[0] == "embed" &&
+            ypath[1].length == 11 && ypath[1] != "videoseries") {
```
lenacohen: One other 11-character embed type sequence we should exclude: "live_stream"

https://webcam.io/docs/live-streaming/embed.html

ghostwords:
Thank you! I made it so that "videoseries" and "live_stream" embeds don't get a hyperlink at all for now.

PR #3218, opened 2026-07-24
Convert nag modal to native dialog element

### LINKED ISSUE #3141: Keyboard navigation not properly trapped within welcome page reminder/critical error/important update notification modal

Capezzuto:
Per the discussion in closed pull request #3206 , this code update converts the nag modal to a native dialog element configured as  a modal, improving semantics and providing focus trapping in the process. A few key points:

- Opens as a modal, using `.showModal()` method, provides focus-trapping
- In keeping with prior behavior, the nag does not dismiss with Esc key, since Esc key closes entire extension popup. Nag modal will persist until user interaction with either the close button or the `button#intro-reminder-btn`
- Scrim opacity is handled through the `::backdrop` pseudo-element, so `div#instruction-outer` has been removed

Fixes #3141

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -86,9 +86,6 @@ function showNagMaybe() {
   }
 
   function _showNag() {
-    nag.showModal();
-    nag.focus(); // work around autofocus not working as expected in Chrome
-
     // Attach event listeners
     $('#fittslaw').on("click", function (e) {
       e.preventDefault();
@@ -103,6 +100,8 @@ function showNagMaybe() {
         window.close();
       });
     });
+
+    nag.showModal();
   }
 
   function _showError(error_text) {
@@ -110,8 +109,8 @@ function showNagMaybe() {
 
     $('#error-message').text(error_text);
 
-    $('#error-text').show().find('a')
-      .addClass('cta-button')
+    let $cta = $('#error-text').show().find('a');
+    $cta.addClass('cta-button')
       .css({
         borderRadius: '3px',
         display: 'inline-block',
@@ -126,7 +125,7 @@ function showNagMaybe() {
     });
 
     nag.showModal();
-    nag.focus();
+    $cta[0].focus();
   }
 
   if (POPUP_DATA.criticalError) {
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -64,8 +64,7 @@ function getPromo(linkRotation) {
 
 /* if they aint seen the comic*/
 function showNagMaybe() {
-  var $nag = $("#instruction");
-  var $outer = $("#instruction-outer");
+  let nag = document.getElementById('instruction');
   let intro_page_url = chrome.runtime.getURL("/skin/firstRun.html");
 
   function _setSeenComic(cb) {
@@ -76,13 +75,11 @@ function showNagMaybe() {
   }
 
   function _hideNag() {
-    $nag.fadeOut();
-    $outer.fadeOut();
+    nag.close();
   }
 
   function _showNag() {
-    $nag.show();
-    $outer.show();
+    nag.showModal();
     // Attach event listeners
     $('#fittslaw').on("click", function (e) {
       e.preventDefault();
@@ -118,8 +115,7 @@ function showNagMaybe() {
       _hideNag();
     });
 
-    $nag.show();
-    $outer.show();
+    nag.showModal();
   }
 
   if (POPUP_DATA.criticalError) {
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -75,7 +75,14 @@ function showNagMaybe() {
   }
 
   function _hideNag() {
-    nag.close();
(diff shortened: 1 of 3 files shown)


src/js/popup.js
```diff
@@ -64,8 +64,7 @@ function getPromo(linkRotation) {
 
 /* if they aint seen the comic*/
 function showNagMaybe() {
-  var $nag = $("#instruction");
-  var $outer = $("#instruction-outer");
+  var nag = document.getElementById('instruction');
```
ghostwords: Nit: Let's use `let` going forward.

src/skin/popup.html
```diff
@@ -64,9 +64,7 @@ <h1 class="i18n_share_title overlay-title"></h1>
 
 <div id="popup-content">
 
-  <div id="instruction-outer">
-  </div>
-  <div id="instruction" role="dialog" aria-modal="true">
+  <dialog id="instruction" aria-modal="true" closedby="none">
```
ghostwords: We don't need to set `aria-modal` ourselves anymore; we should get that already by calling `showModal()`.

src/js/popup.js
```diff
@@ -76,13 +75,11 @@ function showNagMaybe() {
   }
 
   function _hideNag() {
-    $nag.fadeOut();
-    $outer.fadeOut();
```
ghostwords, line 80: Could we add a fade out effect? No big deal if we don't.

Using jQuery's `fadeOut(callback)` pattern and then calling `close()` in the callback seems to produce an unwanted flash when `close()` is invoked. If we can work around the flash, great. Otherwise, maybe this can be achieved with CSS animations?

ghostwords (commented):
Thank you!

>When using [HTMLDialogElement.showModal()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal) to open a `<dialog>`, focus is set on the first nested focusable element. Explicitly indicating the initial focus placement by using the [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus) attribute will help ensure initial focus is set on the element deemed the best initial focus placement for any particular dialog. When in doubt, as it may not always be known where initial focus could be set within a dialog, particularly for instances where a dialog's content is dynamically rendered when invoked, the <dialog> element itself may provide the best initial focus placement.

-- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog

We should either restore previous focus behavior (the dialog itself received focus, I think), or place focus on the call-to-action button. It feels like we want to place keyboard focus on the CTA button.

Capezzuto:
Thanks for the review! These new commits should address the feedback. A brief summary:

- Switched `nag` variable to `let` keyword
- Added `autofocus` attribute to `<dialog>` element, since the content is dynamic, and because I think it will be more useful for screen readers to have the `<dialog>` focused, rather than an individual `<a>` tag.
- Removed redundant `aria-modal="true"`
- Created a fade-out effect that should work cross-browser in the _hideNag() fn. I observed the same behavior with the jQuery `fadeOut()` method's callback that you did. I created a transition effect through adding and removing classes that seems to work without the flash. 
 
Regarding the last bullet point, it seems that pure CSS animations/transitions won't work in FF since the `allow-discrete` value for the `transition-behavior` property is not supported when transitioning between `display:none` and any other display value. I think this would be my preferred method if there were cross-browser support. Maybe at some point in the future it would be good to switch to this approach.

ghostwords:
Ahh, according to https://www.matuzo.at/blog/2023/focus-dialog/

>When you put autofocus on a dialog that contains interactive elements, Safari and Firefox put focus on the dialog, but Chrome puts it on the first focusable element.

ghostwords:
I force-pushed a workaround for the Chrome autofocus bug. I also restored an inadvertently removed `width: calc(100% - 46px);` CSS rule to ensure proper dialog width in all situations.

ghostwords (approved):
Thank you again!

PR #3224, opened 2026-08-04
fix: remove hardcoded direction:ltr from domain rows for RTL locale support (Fixes #3103)

### LINKED ISSUE #3103: Domain toggles are always LTR even in RTL locales

waterWang:
Fixes #3103 - Domain toggles are always LTR even in RTL locales.

Changes:
- Removed direction:ltr hardcoding from .clicker rows
- Converted .clicker to use flexbox layout instead of floats
- Replaced margin-left with margin-inline-start on .switch-container
- Removed direction:ltr from .keyContainer and .key
- Replaced left positioning with margin-inline-start on .key
- Updated responsive media queries consistently

make lint passes.

### DIFF
--- a/src/skin/popup.css
+++ b/src/skin/popup.css
@@ -75,23 +75,25 @@ button {
     overflow: hidden;
     line-height: normal;
     position: relative; /* so that we can abs position child els (removeDomain) */
+    display: flex;
+    align-items: center;
 }
 .clicker:not(#not-yet-blocked-header):not(#non-trackers-header) {
-    direction: ltr;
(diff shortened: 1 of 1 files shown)


ghostwords:
Closing as non-compliant with EFF's AI-generated contributions policy (see [Coding with EFF](https://www.eff.org/about/opportunities/volunteer/coding-with-eff)).

PR #3226, opened 2026-08-06
Add Android-specific graphic with instructions for opening the popup to the Welcome Page

### LINKED ISSUE #3186: Welcome page images do not show Android users how to open the popup
Update the images on the welcome page for Android to accurately show how to open the popup. At minimum, update the "Try opening Privacy Badger's popup" graphic to provide visual instructions. Other images showing what's on the popup could be updated to accurately show where the popup is located, but that's less important since the focus of the images are the popup contents.

Since opening the popup is more complicated on Android and has several steps with different icons across browsers, it could be difficult to make one graphic that works for Firefox AND Edge.

Image

Image

lenacohen:
- Fixes #3186. Addresses the minimum image update. Does not address the other Welcome Page images that show a non-Android browser toolbar, but whose focus is the popup content
- I won't be able to test on an actual Android device until Monday, but until then, here's what the images look like on an mobile-sized viewport:
Screenshot 2026-08-05 at 9 21 14 PM
Screenshot 2026-08-05 at 9 22 42 PM


ghostwords (commented):
Thank you!

We don't want to flip these in RTL locales, right? But I think they will be flipped. Maybe we should stop flipping since it makes the PB icon face the wrong way, and we don't flip any of the other LTR-only graphics.

Also the image will still have alt text that says "Browser toolbar with the Privacy Badger icon circled". I'm OK with leaving that in as a bug to fix for later.

ghostwords:
Not sure why yet, but I don't see the Firefox for Android-specific graphic in Firefox on an Android device.

ghostwords:
Ah, it's a JS error: `Uncaught SyntaxError: import declarations may only appear at top level of a module`

PR #3236, opened 2026-09-16
Validate promo link URLs

ghostwords:
Following up on #3187


src/js/background.js
```diff
@@ -697,17 +697,24 @@ Badger.prototype = {
...
         }
-      } else {
-        // Remove the review promo from rotation when no review URL was defined
-        popupPromos = popupPromos.filter(promo => promo.text !== "popup_review_pb");
       }
 
+      // validate: remove empty and invalid promos
+      popupPromos = popupPromos.filter(promo => {
+        try {
+          if (promo.url && (new URL(promo.url).protocol) == "https:") {
```
lenacohen, line 711: The code LGTM but I'm wondering why checking the protocol is necessary if all of our review links use the https protocol? 
https://github.com/EFForg/privacybadger/blob/f2cd6c23707422b1d47247816c4557c2b4635939/src/js/constants.js#L47..L50

ghostwords: This is in response to what is probably genAI review feedback from an extension store. So in case of an MITM'd connection no `javascript:` URLs are accepted.

lenacohen: Ah I see!

lenacohen (approved):
LGTM!

PR #3237, opened 2026-09-20
Do not drop a cookie that is named after a cookie attribute

arpitjain099:
`skipAttributes` in `parseCookie` matches an attribute by name alone, wherever it appears. A `Cookie:` request header has no attributes, so a name like `path`, `secure` or `expires` there is an ordinary cookie name, and `hasCookieTracking` drops it:

```
parseCookie("path=abc123def456", {noDecode, skipAttributes, skipNonValues})
  -> {}
```

Nothing is left to measure, so `hasCookieTracking` returns false and `_recordPrevalence` is never reached, which keeps that third party out of the count toward the three-first-parties threshold. The same name as the first pair of a `Set-Cookie` behaves the same way. Any of the seven attribute names works, on both the request and the response side.

Two changes: attributes are only skipped after the first pair, since the cookie itself always comes first, and `_extractCookies` now says whether a header can carry attributes at all so a `Cookie:` header is never filtered that way. `Set-Cookie: abc=123; path=/; HttpOnly` still parses to `{abc: '123'}`.

Added two cases beside the existing SameSite ones. I checked them by running `parseCookie` under node rather than the browser suite; eslint is clean.


ghostwords:
Hello, could you talk about the background behind this pull request? What led you to investigate this? What issues does this fix from a user's perspective?

ghostwords:
Could you also review and respond here to the applicable points under the "Policy Regarding AI-Assisted Contributions" subheading on the [Coding with EFF](https://www.eff.org/about/opportunities/volunteer/coding-with-eff) page?

